PR #3, opened 2020-02-06
Eslint

akoushke:


.eslintrc
```diff
@@ -0,0 +1,73 @@
+{
+  "parser": "babel-eslint",
+  "extends": [
+    "@webex/eslint-config-react",
```
lalli-flores: Should we extend from Airbnb's instead of this one?

.eslintrc
```diff
@@ -0,0 +1,73 @@
...
+    "plugin:jsx-a11y/recommended",
+    "plugin:prettier/recommended",
+    "prettier/react"
+  ],
+  "plugins": ["react", "jsx-a11y", "import", "prettier", "react-hooks"],
```
lalli-flores: Should we add the JSDoc plugin?

.eslintrc
akoushke: can we do that after we come up with our standard eslint configuration

.eslintrc
```diff
@@ -0,0 +1,73 @@
...
+        "array": true,
+        "object": true
+      },
+      {
+        "enforceForRenamedProperties": false
+      }
+    ],
+    "react-hooks/rules-of-hooks": "error", // Checks rules of Hooks
+    "react-hooks/exhaustive-deps": "warn", // Checks effect dependencies
+    "require-jsdoc": "warn"
```
lalli-flores: Can remove this since we're adding the plugin

.circleci/config.yml
```diff
@@ -35,8 +35,19 @@ jobs:
           root: .
           paths: node_modules
 
+  test:
+      executor: main-executor
```
lalli-flores:
```suggestion
      executor: node-erbium-executor
```

.eslintrc
```diff
@@ -0,0 +1,72 @@
+{
+  "parser": "babel-eslint",
+  "extends": [
+    "@webex/eslint-config-react",
+    "plugin:jsx-a11y/recommended",
+    "plugin:prettier/recommended",
+    "prettier/react"
+  ],
+  "plugins": ["import", "jsdoc", "jsx-a11y", "prettier", "react", "react-hooks"],
```
lalli-flores: We should also add the `jest` plugin

akoushke: sure

lalli-flores:
Minor thing, if you can lower case the `Add` in your index commit message but approved otherwise

adamweeks:
> Minor thing, if you can lower case the `Add` in your index commit message but approved otherwise

Not to pile on, but is there a reason `Index` is capitalized?

package.json
```diff
@@ -4,7 +4,9 @@
   "main": "dist/webexWidgets.cjs.js",
   "module": "dist/webexWidgets.esm.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "prettier": "prettier \"src/**/!(static)/**/*.*(js|jsx)\"",
```
adamweeks: Let's not use prettier, just the eslint plugin for prettier rules.

.eslintrc
adamweeks: I agree with @lalli-flores, we should use the newly agreed upon eslint config instead of using the soon to be deprecated eslint-config.

PR #4, opened 2020-02-07
Babel

akoushke:


package.json
```diff
@@ -46,5 +46,17 @@
     "eslint-plugin-react-hooks": "^2.3.0",
     "nodemon": "^2.0.2",
     "prettier": "^1.19.1"
+  },
+  "babel": {
+    "presets": [
+      "@babel/preset-env",
+      "@babel/preset-react"
+    ],
+    "plugins": [
```
adamweeks: Please add
"@babel/plugin-proposal-nullish-coalescing-operator"
"@babel/plugin-proposal-optional-chaining"

package.json
```diff
@@ -46,5 +46,17 @@
...
+      "@babel/plugin-proposal-object-rest-spread",
+      "@babel/plugin-proposal-export-default-from",
```
adamweeks: Can you sort these alphabetically?

package.json
```diff
@@ -4,6 +4,7 @@
   "main": "dist/webexWidgets.cjs.js",
   "module": "dist/webexWidgets.esm.js",
   "scripts": {
+    "start": "NODE_ENV=development nodemon src/index.js --exec babel-node",
```
lalli-flores: I don't think this is how we should do our start script. I would leave this out for a different PR once we agree on the repo structure

package.json
```diff
@@ -22,17 +23,48 @@
...
+    "eslint-config-react": "^1.1.7",
     "eslint-plugin-import": "^2.20.1",
     "eslint-plugin-jest": "^23.6.0",
     "eslint-plugin-jsdoc": "^21.0.0",
     "eslint-plugin-jsx-a11y": "^6.2.3",
+    "eslint-plugin-mocha": "^6.2.2",
     "eslint-plugin-prettier": "^3.1.2",
     "eslint-plugin-react": "^7.18.3",
     "eslint-plugin-react-hooks": "^1.7.0",
+    "nodemon": "^2.0.2",
```
lalli-flores: Should probably add this dep separately / in a different PR

package.json
akoushke: that was to check if babel works. yes we dont need it.

lalli-flores: Let's remove it then 👍

akoushke: yes, I'm working on it right now.

PR #5, opened 2020-02-07
Rollup

akoushke:


rollup.config.js
```diff
@@ -0,0 +1,56 @@
...
+      // skip circular dependency warnings from webex library
+      if (warning.code === 'CIRCULAR_DEPENDENCY') return;
+
+			// Use default for everything elsenpm
+      warn(warning);
+    },
+    external: ['bufferutil', 'prop-types', 'react', 'react-dom', 'spawn-sync', 'utf-8-validate'],
+    context: 'null',
+  },
+];
```
lalli-flores: New line

rollup.config.js
```diff
@@ -0,0 +1,56 @@
...
+import json from 'rollup-plugin-json';
+import builtins from 'rollup-plugin-node-builtins';
+
+const output = (name, format) => ({
+  name,
+  file: `dist/webexWidgets.${format}.js`,
+  format,
+  sourcemap: true,
+  globals: {
+		bufferutil: 'bufferutil',
```
lalli-flores: Can you indent this properly?

rollup.config.js
```diff
@@ -0,0 +1,56 @@
...
+    'prop-types': 'PropTypes',
+    react: 'React',
+    'react-dom': 'ReactDOM',
+		'spawn-sync': 'spawnSync',
```
lalli-flores: Can you indent this properly?
Also, what is it for?

rollup.config.js
```diff
@@ -0,0 +1,56 @@
...
+      output('ESMWebexWidgets', 'esm'),
+    ],
+    plugins: [
+      resolve({
+        preferBuiltins: true,
+      }),
+      babel({
+				compact: false,
+				runtimeHelpers: true
+			}),
```
lalli-flores: Can you indent this 3 lines properly?

rollup.config.js
```diff
@@ -0,0 +1,56 @@
...
+      commonJS({
+        // TODO: Remove workaround once fixed in SDK	
+        // explicitly specify unresolvable named exports	
+        namedExports: {'@webex/common': ['deconstructHydraId']},	
+      }),
+      json(),
+			builtins(),
```
lalli-flores: Can you indent this properly?

rollup.config.js
```diff
@@ -0,0 +1,56 @@
...
+    ],
+    onwarn(warning, warn) {
+      // skip circular dependency warnings from webex library
+      if (warning.code === 'CIRCULAR_DEPENDENCY') return;
+
+			// Use default for everything elsenpm
```
lalli-flores: Can you indent this properly?

src/index.js
```diff
@@ -1 +1,3 @@
 import webex from 'webex';
+import WebexSDKAdapter from '@webex/sdk-component-adapter';
```
lalli-flores: Can you add a comment as to you you're adding this?
```suggestion
// Testing bundle works with component libraries
import WebexSDKAdapter from '@webex/sdk-component-adapter';
```

rollup.config.js
```diff
@@ -0,0 +1,52 @@
...
+        preferBuiltins: true,
+      }),
+      babel({
+        compact: false,
+        runtimeHelpers: true,
+      }),
+      commonJS({
+        // TODO: Remove workaround once fixed in SDK
+        // explicitly specify unresolvable named exports
+        namedExports: {'@webex/common': ['deconstructHydraId']},
```
lalli-flores: Is this necessary?

rollup.config.js
```diff
@@ -0,0 +1,52 @@
...
+  name,
+  file: `dist/webexWidgets.${format}.js`,
+  format,
+  sourcemap: true,
+  globals: {
+    bufferutil: 'bufferutil',
+    'prop-types': 'PropTypes',
+    react: 'React',
+    'react-dom': 'ReactDOM',
+    'spawn-sync': 'spawnSync',
```
lalli-flores: What does this do?

rollup.config.js
akoushke: yes

rollup.config.js
akoushke: global dependency from webex module. needs to be externally declared.

adamweeks:
`feat(App): import webex componet system dependencies` “component” is misspelled

rollup.config.js
adamweeks: I’m not familiar with this global dependency, can you link me to it so maybe we can remove it. We don’t want anyone else with the Webex module to have to do this.

rollup.config.js
```diff
@@ -0,0 +1,52 @@
...
+      builtins(),
+    ],
+    onwarn(warning, warn) {
+      // skip circular dependency warnings from webex library
+      if (warning.code === 'CIRCULAR_DEPENDENCY') return;
+
+      // Use default for everything else
+      warn(warning);
+    },
+    external: ['bufferutil', 'prop-types', 'react', 'react-dom', 'spawn-sync', 'utf-8-validate'],
```
adamweeks: Should Webex be external too so that it is brought in through node modules?

rollup.config.js
akoushke: Screen Shot 2020-02-10 at 9 38 25 AM
webex might not utilize this library directly since I cant find inside the repo. However, these warnings disappear when I exclude the webex form the bundling process.

rollup.config.js
akoushke: I dont agree with that.  we shouldn't assume that the consumers always install webex, since we are providing `accessToken` as a the only prop right now. and it could also be an optional prop later.

rollup.config.js
akoushke: not trying to reason my approach. but this is what we have in [sdk-adpter](https://github.com/webex/sdk-component-adapter/blob/master/rollup.config.js#L59) rollup configuration too.

rollup.config.js
adamweeks: Gotcha, I was mistaken on how the external config works on rollup.

rollup.config.js
adamweeks: We need to have a tech debt item to figure out what is causing this. IMO this is a workaround and not a solution.

PR #6, opened 2020-02-07
semantic-release

akoushke:


adamweeks:
Do we need to add the `release.config.js` file in this PR as well?

package.json
```diff
@@ -22,6 +25,14 @@
     "@webex/sdk-component-adapter": "^1.19.3",
     "webex": "^1.80.103"
   },
+  "files": [
```
adamweeks: In order to future proof the repo, especially if we build fonts/images/css, I'd prefer to use a more general grep of the dist folder like:

```js
  "files": [
    "dist"
  ]
```

akoushke: as we discussed about this earlier, the only reason that Lalli wanted to use `files` was to specify the files individually (although we could do that with `.npmignore` too).  If we are planning to only declare the `dist` directory, not sure what the point of having `files` property is. I personally prefer `.npmignore` since all other component system related repositories have been utilizing it and we are following the same publishing tools across all these repositories.

adamweeks: One of @lalli-flores's main arguments FOR the files setting is [this article](https://medium.com/@jdxcode/for-the-love-of-god-dont-use-npmignore-f93c08909d8d). You can see he is exporting the `/lib` file in his preferred config. This gives us the benefit of both npmignore and files: you don't accidentally publish things to npm that you didn't want, as well as making it future proof for any additional files you want in your npm package. 

My main concern around using the files option is that we would need a file to be in our npm package but don't include them causing our npm package to break. This option resolves that concern.

akoushke: however, if we are planning to use `rollup-plugin-serve`, since we need to bundle our index.html with the files too, it will not be a bad idea to specify what exact files we would like publish. But again, we could do that with `.npmignore` as well.

akoushke: my only point here is that `import only 'dist' directory` is as same as `import nothing but 'dist' directory`

lalli-flores: I get @adamweeks concern about potentially missing a file if there is something new to publish, I don't know how frequently we would want to publish something new, though. I would like to have the files for publishing specified individually

PR #7, opened 2020-02-20
Webpack Dev Server

akoushke:


webpack.config.js
```diff
@@ -0,0 +1,80 @@
...
+    'prop-types': 'propTypes',
+  },
+  resolve: {
+    extensions: ['.js', '.jsx'],
+  },
+  module: {
+    rules: [
+      {
+        test: /\.(js|jsx)$/,
+        exclude: /(node_modules|bower_components)/,
```
lalli-flores: We don't use Bower
```suggestion
        exclude: /(node_modules)/,
```

webpack.config.js
```diff
@@ -0,0 +1,80 @@
...
+    open: true,
+    overlay: true,
+    hot: true,
+    port: 1234,
+  },
+  plugins: [
+    new CleanWebpackPlugin({}),
+    new HtmlWebpackPlugin({
+      filename: 'index.html',
+      template: 'src/index.html',
```
lalli-flores: As discussed in the Teams messages, we should probably have this be on a `demo` folder outside `src`

webpack.config.js
adamweeks: You can remove `bower_components`

webpack.config.js
```diff
@@ -0,0 +1,80 @@
+const path = require('path');
+const webpack = require('webpack');
+const {CleanWebpackPlugin} = require('clean-webpack-plugin');
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+
+module.exports = {
+  cache: true,
+  node: {
```
lalli-flores: Why are you overwriting this?

webpack.config.js
```diff
@@ -0,0 +1,80 @@
...
+        ],
+      },
+    ],
+  },
+  devServer: {
+    contentBase: path.resolve(__dirname, './src'),
+    open: true,
+    overlay: true,
+    hot: true,
+    port: 1234,
```
adamweeks: could you add `https: true,` here? Browsers are starting to get very strict on https requirement for local media access.

akoushke:
I can't push to this PR anymore. will create another PR and addressed the comment there.

lalli-flores:
@akoushke next time ask for help. We lose the history otherwise.
Replaced by https://github.com/webex/widgets/pull/7

PR #8, opened 2020-02-20
Webpack Dev Server

akoushke:


demo/App.css
```diff
@@ -0,0 +1,4 @@
+#root {
```
lalli-flores:
```suggestion
#widgets {
```

webpack.config.js
```diff
@@ -0,0 +1,70 @@
...
+          },
+        ],
+      },
+      {
+        test: /\.css$/,
+        use: ['to-string-loader', 'style-loader', 'css-loader'],
+      },
+      {
+        test: /\.scss$/,
+        use: ['to-string-loader', 'style-loader', 'css-loader', 'sass-loader'],
```
lalli-flores: Why are you transforming the CSS/Sass to string?

webpack.config.js
```diff
@@ -0,0 +1,70 @@
...
+        ],
+      },
+    ],
+  },
+  devServer: {
+    contentBase: path.resolve(__dirname, './demo'),
+    open: true,
+    overlay: true,
+    hot: true,
+    port: 1234,
```
lalli-flores: Can you use a different port? This is already used in the [SDK adapter repo](https://github.com/webex/sdk-component-adapter/blob/master/jest-puppeteer.config.js#L4). Maybe `9000`, like the [webpack examples](https://webpack.js.org/configuration/dev-server/#devserver)?

webpack.config.js
```diff
@@ -0,0 +1,70 @@
...
+            },
+          },
+        ],
+      },
+    ],
+  },
+  devServer: {
+    contentBase: path.resolve(__dirname, './demo'),
+    open: true,
+    overlay: true,
```
lalli-flores: Can you make this [show warnings](https://webpack.js.org/configuration/dev-server/#devserveroverlay) as well? Since we're going to use this while developing, we should make sure we also clear out warnings in our code

webpack.config.js
```diff
@@ -0,0 +1,70 @@
...
+    hot: true,
+    port: 9000,
+    stats: 'errors-warnings',
```
lalli-flores: Is this a valid value? https://webpack.js.org/configuration/dev-server/#devserverstats-

PR #9, opened 2020-02-21
Webpack Build - WIP

akoushke:
This is the design that I came up to handle both prod and dev bundling process using webpack. feel free to play around with it and leave an questions/feedbacks you might have. 

our bundle does not work with CRA ATM.


webpack/webpack.prod.config.js
```diff
@@ -0,0 +1,25 @@
...
+      filename: `webexWidgets.${affix}.js`,
+      library: 'WebexWidgets',
+      libraryTarget: format,
+    },
+  };
+
+  return Object.assign(prodConfig, commonConfig);
+}
+
+module.exports = [createConfig('cjs', 'commonjs'), createConfig('umd', 'umd'), createConfig('esm', 'var')];
```
akoushke: I think `var` combines all formats together. not sure which format is only for `esm`.  more info here: https://webpack.js.org/guides/author-libraries/#add-librarytarget

adamweeks: > variable: as a global variable made available by a script tag (libraryTarget:'var').

Definitely not the option we're looking for, this would make it so you could do `window.WebexWidgets`.

It looks like webpack doesn't support esm building until webpack v5 according to https://github.com/webpack/webpack/issues/2933

I believe this was the reason we went with rollup for esm in the original widgets.

akoushke:
https://github.com/webex/widgets/pull/10 will take over this PR

PR #10, opened 2020-02-26
Exclude webex from the bundle

akoushke:
I have the tested the `esm` bundle on CRA already.


akoushke:
should wait for https://github.com/webex/widgets/pull/10

PR #11, opened 2020-02-26
initial Husky configuration

akoushke:


package.json
```diff
@@ -97,5 +101,26 @@
...
+          "lower-case",
+          "pascal-case"
+        ]
+      ]
+    }
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run test:eslint",
+      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
```
lalli-flores: A bit out of scope, but is there a reason why we're not running tests `pre-push`?

src/index.js
```diff
@@ -1,4 +1 @@
-// Testing bundle works with webex and component libraries
-import webex from 'webex';
-import WebexSDKAdapter from '@webex/sdk-component-adapter';
```
lalli-flores: Why are this changes here? I think you need to rebase your branch from master again

package.json
```diff
@@ -40,12 +40,15 @@
...
     "@babel/preset-react": "^7.8.3",
+    "@commitlint/cli": "^8.3.5",
+    "@commitlint/config-conventional": "^8.3.4",
     "@semantic-release/changelog": "^5.0.0",
     "@semantic-release/git": "^9.0.0",
     "babel-eslint": "^10.0.3",
     "babel-loader": "^8.0.6",
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "clean-webpack-plugin": "^3.0.0",
+    "conventional-changelog-angular": "^5.0.6",
```
lalli-flores: Why are you installing this explicitly? This already gets installed as part of `@commitlint` packages

akoushke: ohhh thats why package.locked didnt change....

akoushke: wait,are you sure? it didnt add it for me. though I didnt install @commitlint. I installed cli and config and then this.

taymoork2:
@lalli-flores is right chore is what you'd use here

package.json
akoushke: with github already running the tests for every push, i find it a bit pointless to have this script. with that being said, I don't mind adding a `pre-push` script. but not yet at least. maybe after adding some tests to this repo.

lalli-flores: I don’t think it’s pointless for the `pre-push` commit to run. Having the developer running tests and making sure they work before pushing to Github (specially if a PR is open) would save us from running unnecessary build in our CI. Freeing up resources is important as the project grows. It would be a good standard across all our projects

akoushke: will add it when we come with the testing environment.

package-lock.json
```diff
@@ -1141,33 +1293,53 @@
         "base64-js": "^1.3.0"
       }
     },
+<<<<<<< Updated upstream
```
lalli-flores: There are Git conflict markers on `package-lock.json`

PR #16, opened 2020-03-20
docs(demo): add token and destination input

adamweeks:
This PR expands upon the existing demo page and allows the user to input an access token and widget destination without having to modify source code.

image

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-129602


demo/App.jsx
```diff
@@ -1,11 +1,62 @@
...
+        </SidebarBody>
+      </Sidebar>
+      <div className="content">
+        <div className="section">
+          <h3>Webex Widgets Access Token</h3>
+          <h5>Webex Widgets require an access token to identify the current user.</h5>
+          <h5>
+            You can get an access token from <a href="https://developer.webex.com">developer.webex.com</a>
+          </h5>
+          <form style={{padding: '1em'}}>
```
taymoork2: I'm either or on moving style into the css and using a class instead

adamweeks: I can move it to a class for consistency. This was just lazy 🤣

demo/App.jsx
```diff
@@ -1,11 +1,62 @@
...
+    setTokenInput('');
+    setToken();
+  };
+
+  return (
+    <div className="main">
+      <Sidebar withIcons={false}>
+        <SidebarBody>
+          <SidebarNav title="Webex Widgets">
+            <SidebarNavItem title="WebexMeeting" />
```
lalli-flores:
```suggestion
            <SidebarNavItem title="Webex Meeting Widget" />
```

demo/App.jsx
```diff
@@ -1,11 +1,62 @@
...
+          </SidebarNav>
+        </SidebarBody>
+      </Sidebar>
+      <div className="content">
+        <div className="section">
```
lalli-flores: There is a semantic HTML element, [`section`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section). I think this also represents the intention in a semantic way 😬

demo/App.jsx
```diff
@@ -1,11 +1,62 @@
...
+            <button onClick={() => setToken(tokenInput)} disabled={token}>
+              Save Token
+            </button>
+            <button onClick={clearToken} disabled={!token}>
+              Clear Token
+            </button>
+          </form>
+        </div>
+        <div className="section">
+          <h2>WebexMeetingWidget</h2>
```
lalli-flores: Can you use an `h3` to keep the size consistency with the previous section?
```suggestion
          <h3>Webex Meeting Widget</h3>
```

demo/App.css
```diff
@@ -1,4 +1,27 @@
...
-  height: 100%;
+}
+
+.form {
+  padding: 1em;
+}
+
+.section {
+  background-color: white;
+  width: 95%;
```
lalli-flores: Is this arbitrary? I would prefer to add some more padding instead

demo/App.css
```diff
@@ -1,4 +1,27 @@
...
+  margin: 1em auto 0;
+  padding: .5em 1em;
+  box-shadow: rgba(0, 0, 0, 0.12) 0px 1px 6px, rgba(0, 0, 0, 0.12) 0px 1px 4px;
+  border-radius: 2px;
```
lalli-flores: Can you change to `rem`? Actually, all units?

demo/WebexMeetingWidgetDemo.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+WebexMeetingWidgetDemo.propTypes = {
+  token: PropTypes.string,
+};
+
+WebexMeetingWidgetDemo.defaultProps = {
+  token: '',
+};
+
+export default WebexMeetingWidgetDemo;
```
lalli-flores: All exports are done in the declaration, any reason for this to be done separately?

demo/WebexMeetingWidgetDemo.jsx
```diff
@@ -0,0 +1,46 @@
...
+import {WebexMeetingWidget} from '../src';
+
+function WebexMeetingWidgetDemo({token}) {
+  const [destination, setDestination] = useState('');
+  const [displayWidget, setDisplayWidget] = useState(false);
+
+  const displayButtonEnabled = token && destination && !displayWidget;
+
+  return (
+    <div>
```
lalli-flores: Maybe a fragment is a better choice here
```suggestion
    <React.Fragment>
```

demo/WebexMeetingWidgetDemo.jsx
```diff
@@ -0,0 +1,46 @@
...
+      <Input
+        htmlId="destination"
+        label="Widget Destination (email, room, sip, etc)"
```
lalli-flores: Let's just be explicit on what we support
```suggestion
        label="Widget Destination (email, person ID, room ID, SIP)"
```

adamweeks:
@lalli-flores ready for another round

demo/App.scss
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  section {
+    background-color: white;
+    margin: 1rem auto 0;
+    padding: .5rem 1rem;
+    box-shadow: 0 .0625rem .125rem 0 #c6c7ca;
+    border-radius: .25rem;
+  }
+}
```
lalli-flores: So I think the reason you had `main` is because the sidebar would not grow when the content did. This goes to using un-semantic markup for the sake of styling as the `widget-demo` wrapper already exists. So my solution would be to make the content scrollable, instead of the whole page.
Here's my CSS (includes all the other suggestions on markup)
```CSS
#widgets-demo {
  display: flex;
  height: 100%;
  width: 100%;

  .content {
    width: 100%;
    height: 100%;
    padding: 1rem;

    overflow: auto;
  }

  form {
    margin: 1rem;
  }

  section {
    margin: 1rem 0;
    padding: 0.5rem 1rem;

    background-color: white;
    box-shadow: 0 0.0625rem 0.25rem 0 #C6C7CA;
    border-radius: 0.25rem;
  }
}

File: demo/App.jsx [hunk 11]
```diff
@@ -1,11 +1,62 @@
...
+  const [token, setToken] = useState();
+
+  const clearToken = () => {
+    setTokenInput('');
+    setToken();
+  };
+
+  return (
+    <div className="main">
+      <Sidebar withIcons={false}>
```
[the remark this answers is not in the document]
adamweeks: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-122360

demo/App.scss
```diff
@@ -0,0 +1,26 @@
...
+
+  form {
+    padding: 1rem;
+  }
+
+  section {
+    background-color: white;
+    margin: 1rem auto 0;
+    padding: .5rem 1rem;
+    box-shadow: 0 .0625rem .125rem 0 #c6c7ca;
```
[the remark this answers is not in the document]
adamweeks: I am using the style from the momentum components page https://momentum.design/components/overview

I will add the leading 0's and capitalization of the hex codes. If this is something you want going forward I suggest finding a css linter and adding the rule so we don't have this back and forth on css.

demo/WebexMeetingWidgetDemo.jsx
```diff
@@ -0,0 +1,44 @@
...
+        onChange={(event) => setDestination(event.target.value)}
+        placeholder="Widget Destination"
+        readOnly={displayWidget}
+        type="text"
+        value={destination}
+      />
+      <button disabled={!displayButtonEnabled} onClick={() => setDisplayWidget(true)}>
+        Display Meeting Widget
+      </button>
+      <button disabled={!displayWidget} onClick={() => setDisplayWidget(false)}>
```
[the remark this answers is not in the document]
adamweeks: Correct, since we don't support updating the destination live, the widget has to be torn down before updating.

adamweeks:
@lalli-flores done

PR #18, opened 2020-03-25
Build and persist docs changes to repo for gh-pages usage

adamweeks:
This PR takes the sample demo that is run via `npm start` and compiles it into a webpage that can be served via github pages.


webpack.config.js
```diff
@@ -44,7 +48,7 @@ module.exports = {
           {
             loader: 'file-loader',
             options: {
-              outputPath: 'assets/',
+              name: 'assets/[name].[ext]',
```
adamweeks: This change was so the assets don't get renamed to a hash every time since we don't wipe the docs folder.

.circleci/config.yml
```diff
@@ -76,9 +87,16 @@ workflows:
...
+      - docs:
+          requires:
+            - test
+          filters:
+            branches:
+              only: master
       - release:
           requires:
             - build
+            - docs
```
adamweeks: This flow matches the components flow: https://github.com/webex/components/blob/master/.circleci/config.yml#L101

adamweeks:
@InteractiveTimmy 

> Looks good, but, I wanted to confirm if we are building to `/docs` or `/docs/subfolder`. I'd prefer the latter as it leaves provisioning for other GH pages.
> 
> Non-blocking comments, but I'd like to provision spacing for other types of documentation within the `./docs` folder to statically serve to ghpages.

We don't have any documentation being generated, so instead of providing an empty page at `/docs` with a link to `/docs/demo` we opted to put the demo into `/docs` for now.

If we have features that will go into the docs folder, we can update at that point.

> Based on the code changes, It appears that general docs build here: `./docs`. and the Demo builds here: `./demo`.
> 

Demo builds to the docs folder but the code is served from the demo folder. Think of `/demo` as `/src` and `/docs` as `/dist` specifically for the widget demo.

InteractiveTimmy:
> @InteractiveTimmy
> 
> > Looks good, but, I wanted to confirm if we are building to `/docs` or `/docs/subfolder`. I'd prefer the latter as it leaves provisioning for other GH pages.
> > Non-blocking comments, but I'd like to provision spacing for other types of documentation within the `./docs` folder to statically serve to ghpages.
> 
> We don't have any documentation being generated, so instead of providing an empty page at `/docs` with a link to `/docs/demo` we opted to put the demo into `/docs` for now.
> 
> If we have features that will go into the docs folder, we can update at that point.
> 
> > Based on the code changes, It appears that general docs build here: `./docs`. and the Demo builds here: `./demo`.
> 
> Demo builds to the docs folder but the code is served from the demo folder. Think of `/demo` as `/src` and `/docs` as `/dist` specifically for the widget demo.

This makes sense. Since the widgets are designed as more of a consumer product rather than an SDK. Everything looks good to me!

lalli-flores:
Looks good, thanks for addressing

PR #27, opened 2021-01-21
Update to flow when leaving a meeting

lalli-flores:
When a user leaves the meeting, a message stating that the meeting has ended is displayed, instead of sending back the user to the interstitial page.
This prevents issues with meeting controls getting out of sync when the meeting no longer exists, and simplifies the workflow for applications embedding the widget.

Other updates:
- Updating node version and node sass, the demo app wasn't loading as there was a mismatch between the node sass version in the dev dependencies and the node version of the project

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,10 +2,10 @@ version: 2.1
 
 # reusable environment for all jobs
 executors:
-  node-erbium-executor:
+  node-executor:
     working_directory: ~/widgets
     docker:
-      - image: circleci/node:erbium-browsers
+      - image: circleci/node:fermium-browsers
 
 # reusable command for all jobs
 commands:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -2,10 +2,10 @@ version: 2.1
 
 # reusable environment for all jobs
 executors:
-  node-erbium-executor:
+  node-fermium-executor:
```
adamweeks, line 5: Could we just name the executor `node-executor` so we don't have to update it in multiple places for every version? Just have the executor running the proper image, like the change on line 8.

adamweeks (approved):
Nonblocking comment on the executor name. Otherwise looks good. Thanks for including the details of what the component version bump would do for the widget!

.circleci/config.yml
lalli-flores: good point, will do 👍🏼

taymoork2 (approved):

PR #30, opened 2021-04-07
refactor(WebexMeeting): create meeting object in the widget

maloun96:
As refactoring in this PR webex/components#193. We should update the widget to get meetingID and send it to the child.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-223898

### DIFF
--- a/src/widgets/WebexMeeting/WebexMeeting.jsx
+++ b/src/widgets/WebexMeeting/WebexMeeting.jsx
@@ -25,27 +25,38 @@ export default class WebexMeetingWidget extends Component {
     });
     this.state = {
       adapterConnected: false,
+      meeting: null,
     };
     this.adapter = new WebexSDKAdapter(webex);
+    this.meetingSubscription = null;
   }
 
   async componentDidMount() {
     await this.adapter.connect();
     // Once adapter connects, set our app state to ready.
     this.setState({adapterConnected: true});
+
+    this.meetingSubscription = this.adapter.meetingsAdapter
+      .createMeeting(this.props.meetingDestination)
+      .subscribe((meeting) => this.setState({meeting}));
   }
 
   async componentWillUnmount() {
+    if (this.meetingSubscription) {
+      this.meetingSubscription.unsubscribe();
+    }
     // On teardown, disconnect the adapter.
     await this.adapter.disconnect();
   }
 
   render() {
+    const {adapterConnected, meeting} = this.state;
+
     return (
       <div className="meeting-widget">
-        {this.state.adapterConnected ? (
+        {adapterConnected && meeting ? (
           <WebexDataProvider adapter={this.adapter}>
-            <WebexMeeting meetingDestination={this.props.meetingDestination} />
+            <WebexMeeting meetingID={meeting.ID} />
           </WebexDataProvider>
         ) : (
           <Spinner />


maloun96:
@lalli-flores

src/widgets/WebexMeeting/WebexMeeting.jsx
```diff
@@ -25,27 +25,38 @@ export default class WebexMeetingWidget extends Component {
...
   }
 
   async componentDidMount() {
     await this.adapter.connect();
     // Once adapter connects, set our app state to ready.
     this.setState({adapterConnected: true});
+
+    this.meetingSubscription = this.adapter.meetingsAdapter
+      .createMeeting(this.props.meetingDestination)
+      .subscribe((meeting) => this.setState({meeting}));
```
lalli-flores, line 41: `createMeeting` returns a [meeting objecthttps://github.com/webex/component-adapter-interfaces/blob/master/src/MeetingsAdapter.js#L20] which doesn't have a `meeting` property

src/widgets/WebexMeeting/WebexMeeting.jsx
```diff
@@ -25,27 +25,38 @@ export default class WebexMeetingWidget extends Component {
...
     this.adapter = new WebexSDKAdapter(webex);
+    this.meetingSubscription = null;
   }
 
   async componentDidMount() {
     await this.adapter.connect();
     // Once adapter connects, set our app state to ready.
     this.setState({adapterConnected: true});
+
+    this.meetingSubscription = this.adapter.meetingsAdapter
```
lalli-flores, line 39: Let's follow the observable name pattern
```suggestion
    this.meeting$ = this.adapter.meetingsAdapter
```

src/widgets/WebexMeeting/WebexMeeting.jsx
```diff
@@ -25,27 +25,38 @@ export default class WebexMeetingWidget extends Component {
...
+      .createMeeting(this.props.meetingDestination)
+      .subscribe((meeting) => this.setState({meeting}));
   }
 
   async componentWillUnmount() {
+    if (this.meetingSubscription) {
+      this.meetingSubscription.unsubscribe();
```
lalli-flores, line 46: It won't harm to do this, but `createMeeting` emits once and completes so there is no need to unsubscribe

lalli-flores (changes requested):

src/widgets/WebexMeeting/WebexMeeting.jsx
cipak: actually, the meeting object is not destructured here, it's received as a whole and put on a state property

src/widgets/WebexMeeting/WebexMeeting.jsx
cipak: we don't really need the observable itself here, we store just the subscription to it (so we can unsubscribe later). I was thinking of naming it `meetingSub` for brevity, but then there is a potential confusion with the term "subject".

I also saw a pattern of having subscriptions in a component indexed by name in a `this.subscriptions` object, it's a bit more generic than we currently need here but it could work:

```js
// constructor
this.subscriptions = {};

// didMount
this.subscriptions.meeting = obs.subscribe(...);

// willUnmount
for(let sub of Object.values(this.subscriptions)) {
  sub.unsubscribe();
}
```

src/widgets/WebexMeeting/WebexMeeting.jsx
cipak: @lalli-flores This is due to my confusion between `createMeeting` and `getMeeting`.

You're correct, `createMeeting` emits a single message, while `getMeeting` emits a new meeting on each meeting event, keeping the subscriber up to date with the meeting object. The widget will need this updating to respond to status changes (eg: when the user leaves the meeting, the widget should hide the roster).

Can we return from `createMeeting` the same thing that `getMeeting` does? It seems that whenever a meeting is created, `getMeeting` is called immediately anyway.

If you prefer to keep `createMeeting` just for meeting creation (and it makes sense to have it so), then I will kindly ask @maloun96 to re-add the `.pipe(getMeeting)` line I asked him to remove yesterday:)

src/widgets/WebexMeeting/WebexMeeting.jsx
lalli-flores: My bad, I only saw the first line and didn't realize the rest of the operators in the following lines 😅

Actually, the most efficient way to dispose of all subscriptions is to use the `add()` operator of a `Subscription` and unsubscribe from the subscription that hold them all.
See https://rxjs-dev.firebaseapp.com/guide/subscription and how we use it [in our integration tests](https://github.com/webex/sdk-component-adapter/blob/master/src/RoomsSDKAdapter.integration.test.js#L88)

src/widgets/WebexMeeting/WebexMeeting.jsx
lalli-flores: I can see why it can be confusing, however, I would like to keep `createMeeting` strictly about meeting creation. I know there are use cases out there where you may create a meeting, but not join it right away.

So yes, you would need to re-add .pipe(getMeeting) 😜

src/widgets/WebexMeeting/WebexMeeting.jsx
lalli-flores: You're right, this is `setState` 🤦🏻‍♀️, no hooks here. Should you make it a shallow copy though? I don't want to run into state issues because React is not picking up the change within the object

lalli-flores:
@cipak @maloun96 is this PR still necessary or can it be closed?

PR #35, opened 2021-04-20
build(rollup): add assets to the build

cricku:
Copy the assets to the `assets` folder in the build directory and resolve the paths in the CSS file.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-208853


rollup.config.js
```diff
@@ -33,7 +34,16 @@ export default [
       }),
       commonJS(),
       json(),
-      postcss({extract: 'dist/webexWidgets.css'}),
+      postcss({
+        extract: 'dist/webexWidgets.css',
+        minimize: true,
+        plugins: [
+          copyAssets({ base: 'dist/assets' })
```
lalli-flores, line 41: Where/when is this `assets` directory generated?

cricku: The current problem is that the `webexWidgets.css` file contains relative paths to assets (fonts, images), for example `../../fonts/fontName.woff` but none were exported where the package is generated.

The `copyAssets` plugin copies all the assets that he founds in the `webexWidgets.css` file in the `dist` folder (te bo more clear I placed all the assets in the `assets` folder) and rewrites all the paths from the CSS file to point to the location of the new asset. After package build we will have this structure:

```
dist
└── assets
    ├── fonts
    └── images
└── webexWidgets.css
└── webexWidgets.esm.js
```
All the things are happening at the build time when the `build` script is executed.

package.json
```diff
@@ -76,6 +76,7 @@
     "html-webpack-plugin": "^3.2.0",
     "husky": "^4.2.3",
     "node-sass": "^4.14.1",
+    "postcss-copy-assets": "^0.3.1",
```
lalli-flores, line 79: Is this this the only package out there? This hasn't been touched in over 3 years. Dependencies we add need to be actively maintained, otherwise we are accountable and get bombarded with security vulnerabilities from InfoSec.

rollup.config.js
```diff
@@ -46,4 +56,3 @@ export default [
     context: null,
   },
 ];
-9;
```
lalli-flores, line 49: Why is the EOF line removed?

package.json
cricku: I didn't notice that the package is not maintained anymore. I replace the package with the `postcss-url` which does approximately the same thing.

rollup.config.js
cricku: There was a weird `9;` added on the end line (i think it was introduced by someone by mistake). My intention was to just remove the unwanted code from the last line. Anyway, I can't figure out why git doesn't pick the new line at the end of the file cause VSCode adds it on file save.

package.json
lalli-flores: Thanks for finding an alternative

rollup.config.js
lalli-flores: Okay, thanks for cleaning it up!

lalli-flores (approved):

PR #61, opened 2021-05-10
test(e2e): run e2e tests on local env

alinasuciu:
Add the possibility to run e2e tests on local environment  
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-231402


webpack.config.js
```diff
@@ -63,7 +63,6 @@ module.exports = function(env, argv) {
       argv.mode === 'development'
         ? {
             contentBase: path.resolve(__dirname, './demo'),
-            open: true,
```
cipak, line 66: does this mean that running `npm start` won't automatically open the demo page in the browser anymore?

cipak: To keep this, we can use the `--open` cli option of webpack serve https://github.com/webpack/webpack-cli/blob/master/SERVE-OPTIONS.md

```json
  "scripts": {
    "demo:serve": "webpack-dev-server demo/App.jsx --mode development",
    "demo:build": "webpack demo/App.jsx --mode production",
    "start": "npm run demo:serve -- --open",
    "test:e2e": "cross-env START_SERVER_AND_TEST_INSECURE=1 start-server-and-test demo:serve https-get://localhost:9000 test:e2e",
    "test:eslint": "eslint src/"
  },
```

cipak: And please leave `open: false` so that the intention not to open a browser is explicit.

cipak (approved):

lalli-flores (commented):
Not sure I'm thrilled with this solution. `cross-env` is actually not maintained anymore, so maybe we can just include the env variable in the env files?
I need to look a little into `start-server-and-test` to determine if it's a dependency we want to add. have you tried the static server from webdriver.io?

wdio.conf.js
```diff
@@ -120,7 +120,15 @@ exports.config = {
...
   // 1) You have the minimum required Java version for your OS - https://github.com/vvo/selenium-standalone/blob/master/docs/java-versions.md
   // 2) You have installed the browsers that are defined in capabilities (e.g. Chrome, Firefox, MicrosoftEdge)
-  services: ['selenium-standalone'],
+  services: [
+    ['static-server', {
+      folders: [
+          { mount: '/', path: './docs' },
+        ]
+      },
+      'selenium-standalone'
```
cipak, line 129: i think `selenium-standalone` is a separate service and needs to be outside of `static-server`'s array

wdio.conf.js
```diff
@@ -120,7 +120,15 @@ exports.config = {
   // To run the tests locally you need to ensure that:
   // 1) You have the minimum required Java version for your OS - https://github.com/vvo/selenium-standalone/blob/master/docs/java-versions.md
   // 2) You have installed the browsers that are defined in capabilities (e.g. Chrome, Firefox, MicrosoftEdge)
-  services: ['selenium-standalone'],
+  services: [
+    ['static-server', {
+      folders: [
+          { mount: '/', path: './docs' },
```
cipak, line 126: indentation

alinasuciu:
> Not sure I'm thrilled with this solution. `cross-env` is actually not maintained anymore, so maybe we can just include the env variable in the env files?
> I need to look a little into `start-server-and-test` to determine if it's a dependency we want to add. have you tried the static server from webdriver.io?
I changed the implementation and now we are using static server from webdriverio. Thanks for the hint, I saw that they have this feature but I wasn't sure it would be what we need since we don't have static files. So in order for this to work and the tests to relay on the latest updates it's important to run the build before or I can also add the build command before the test command inside the script.

lalli-flores:
@alinasuciu , awesome! Glad it worked easily. A couple things:
* your commit should be of [`test` type](https://github.com/webex/components/blob/master/CONTRIBUTING.md#type)
* if building is necessary, then please do add the building script before starting the tests
* I'm ok with adding `docs` to `.gitignore`, but please do so in a separate commit. Also make sure to update the CI script for master updates to force add the docs folders since now it will be ignored

alinasuciu:
> @alinasuciu , awesome! Glad it worked easily. A couple things:
> 
> * your commit should be of [`test` type](https://github.com/webex/components/blob/master/CONTRIBUTING.md#type)
> * if building is necessary, then please do add the building script before starting the tests
> * I'm ok with adding `docs` to `.gitignore`, but please do so in a separate commit. Also make sure to update the CI script for master updates to force add the docs folders since now it will be ignored


@lalli-flores , I added the build script before the test and changed the commit label. Regarding the docs file in gitignore, I'll create a new task for it and a new pr. It's important to have this one merged.

PR #72, opened 2021-05-14
fix(WebexMeeting): fix e2e tests

karinasigartau0798:
In order to align with the changes made in WebexMeeting widget, it was necessary to update the name of the class that stylizes the widget. Additionally  it was necessary to give up to the nesting if structure from afterAll, because now the creation of a room no longer depends on the creation of a user.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-233689

### DIFF
--- a/tests/WebexMeeting.e2e.js
+++ b/tests/WebexMeeting.e2e.js
@@ -37,10 +37,11 @@ describe('Meeting Widget', () => {
 
   afterAll(() => {
     MeetingPage.unloadWidget();
+    if (room) {
+      browser.call(() => sdk.rooms.remove(room).catch(console.error));
+    }
+    
     if (user) {
-      if (room) {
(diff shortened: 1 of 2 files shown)


cipak (approved):

lalli-flores (commented):
Since this is only affecting tests, please use `test` commit type. No need to trigger a release

lalli-flores (approved):
Next time let's be more specific about what is getting fixed. By looking at the commit alone I would not be able to tell what really changed

karinasigartau0798:
> Next time let's be more specific about what is getting fixed. By looking at the commit alone I would not be able to tell what really changed

Yes you're right. I thought that for this PR it is better to make a more detailed description, because I didn't find a commit message concise enough to cover all the aspects. Can you please give me a recommendation regarding to which message would have been more appropriate here? I want to know this for future PRs that will be of this type.

PR #74, opened 2021-05-20
build(workflow): remove docs folder from git

lungancatalin:
Remove docs folder from git
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-235221


lalli-flores (changes requested):
You can ignore the `docs` folder but don't remove the files, otherwise the sample app will break

lungancatalin:
> You can ignore the `docs` folder but don't remove the files, otherwise the sample app will break

Hi @lalli-flores,

Aren't docs files automatically generated? If we leave them there, won't there be stale at the next merge? From what I know,
all of these files are automatically generated after each release, that means that folder will be always there. 

Thank you! :)

lalli-flores:
@lungancatalin They are only generated when there is a release, not for every merge. The `docs` folder must always exist because that is where Github pages serve the contents of our sample page.

PR #104, opened 2021-08-04
feat(WebexMeeting): remove control bar and roster from widget

CernaianuMihai:
https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-255717


src/widgets/WebexMeeting/WebexMeeting.jsx
```diff
@@ -40,23 +33,7 @@ class WebexMeetingWidget extends Component {
     } else if (!audioPermission || !videoPermission || !meeting.ID) {
```
cipak, line 33: it's ok to send an empty meeting id to the meeting component, please remove `!meeting.ID` from this condition

CernaianuMihai: done

PR #116, opened 2021-09-08
build(webpack): add possibility to use widget without react dependency

alinasuciu:
Add possibility to use widget without react as dependency

### DIFF
--- a/cdn/index.js
+++ b/cdn/index.js
@@ -1,0 +1,8 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import WebexMeetingWidget from '../src/widgets/WebexMeeting/WebexMeeting';
+
+window.WebexMeetingWidget = (node, props) => {
+  ReactDOM.render(<WebexMeetingWidget {...props}/>, node);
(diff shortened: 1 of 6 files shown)


cdn/index.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import WebexMeetingWidget from '../src/widgets/WebexMeeting/WebexMeeting';
+ 
+function init(node, meetingDestination, accessToken){
```
cipak, line 6: to make room for future widgets, plus adding props to the widget, let's use:
```js
window.WebexMeetingWidget = (node, props) {
  ReactDOM.render(...);
}
``

lalli-flores (changes requested):
We use [rollup](https://github.com/webex/widgets/blob/master/rollup.config.js) to create library bundles. could you please configure the tooling such that we can generate the CDN bundle via Rollup?

lalli-flores:
We will have to revisit this work. Closing for now

PR #128, opened 2021-09-29
feat(WebexMeeting): add usage metrics for meetings widget

alinasuciu:
Add usage metrics for meetings widget.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-270813


.env.default
```diff
@@ -1,3 +1,5 @@
+WEBEX_WIDGETS_VERSION=1.14.0
```
lalli-flores, line 1: How is this variable updated?

src/widgets/WebexMeeting/WebexMeeting.jsx
```diff
@@ -61,6 +61,12 @@ WebexMeetingWidget.defaultProps = {
 export default withAdapter(withMeeting(WebexMeetingWidget), (props) => {
   const webex = new Webex({
     credentials: props.accessToken,
+    ...(process.env.NODE_ENV === 'production'
```
lalli-flores, line 64: Help me think this through, is this step processed at build time? `process` won't be on the final bundle right? Because this doesn't exist on a browser.

PR #218, opened 2024-04-12
docs(contributing-guide): Update CONTRIBUTING.md

Kesari3008:
Updating the missing step for linking dependent repo's locally


sreenara (changes requested):
We haven't mentioned anywhere in the contributing guide about how to link the repos locally. Can we have a paragraph about that along with steps and commands?

sreenara (approved):

sreenara (changes requested):
Please fix the title according to convention so we don't break any semantic release

sreenara (approved):

PR #222, opened 2024-07-09
fix(tooling): upgraded node version from 14 to 20

Kesari3008:
Upgrading node version from 14 to 20.

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-501675 

Tests Performed:
1. Joined meeting as a licensed user.
2. Guest joining a meeting with a password and captcha. 
Screenshot 2024-07-18 at 10 47 07 AM

3. Guest joining the meeting as host by entering host key. 
Screenshot 2024-07-18 at 10 47 17 AM
4. Joined meeting using SIP destination, personID and meeting link.


.circleci/config.yml
```diff
@@ -27,7 +27,7 @@ jobs:
             - dependency-cache-{{ checksum "package-lock.json" }}
       - run:
           name: Install Dependencies
-          command: npm i
+          command: npm i --force #Remove force once webex version is upgraded in sdk-component-adapter
```
sreenara, line 30: Isn't the version upgraded in sdk-component-adapter? Which version are we referring to here?

sreenara (commented):
Have a question about the change in the pipeline config.

.circleci/config.yml
Kesari3008: This is Webex version mismatch which fails the npm install

sreenara (approved):

.circleci/config.yml
```diff
@@ -5,7 +5,7 @@ executors:
   node-executor:
     working_directory: ~/widgets
     docker:
-      - image: circleci/node:fermium-browsers
+      - image: cimg/node:20.13.1-browsers
```
Shreyas281299, line 8: Screenshot 2024-07-18 at 1 04 42 PM
fermium was node-14. I think iron would be 20. Do you think we use iron here instead of the version.

Shreyas281299 (approved):
Just 1 non blocking ask

.circleci/config.yml
Kesari3008: I tried using iron, but it didn't work. I couldn't find circle ci image name with node:iron-browsers

PR #229, opened 2024-10-04
fix: updated sdk-component-adapter to latest to fix joinMeeting

rsarika:
Issue: Not showing password popup when click join meeting after dismissing initial popup.
Fix: added condition to check passwordRequired and empty options to updateMeeting with passwordRequired true

https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-557507 

Screenshot 2024-09-30 at 8 31 25 PM
Screenshot 2024-09-30 at 8 31 32 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   ],
   "dependencies": {
     "@webex/component-adapter-interfaces": "^1.30.5",
-    "@webex/sdk-component-adapter": "1.112.7",
+    "@webex/sdk-component-adapter": "1.112.9",
     "webex": "^2.59.3",
     "@webex/components": "1.275.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -32,7 +32,7 @@
   ],
   "dependencies": {
     "@webex/component-adapter-interfaces": "^1.30.5",
-    "@webex/sdk-component-adapter": "1.112.7",
+    "@webex/sdk-component-adapter": "1.112.9",
```
Kesari3008, line 35: We should also have the package-lock.json updated for this change. Please check

sreenara (approved):

PR #231, opened 2024-10-06
fix(webex): upgrading new version to 2.x version

Parimala032:


webpack.config.js
```diff
@@ -25,6 +25,7 @@ module.exports = function(env, argv) {
         "util": require.resolve("util/"),
         "url": require.resolve("url/"),
         "vm": require.resolve("vm-browserify"),
+        "querystring": require.resolve('querystring-es3'),
```
sreenara, line 28: Please add a comment for why this change is needed.

package.json
```diff
@@ -32,8 +32,9 @@
   ],
   "dependencies": {
     "@webex/component-adapter-interfaces": "^1.30.5",
-    "@webex/sdk-component-adapter": "1.112.9",
-    "webex": "^2.59.3",
+    "@webex/sdk-component-adapter": "1.112.10",
+    "querystring-es3": "^0.2.1",
```
sreenara, line 36: Should this be a dev dependency?

sreenara (changes requested):

webpack.config.js
Parimala032: Yeah @sreenara 
Webpack 5 no longer includes Node.js polyfills by default, which is causing the error for the querystring module. we need to add a fallback configuration for querystring in webpack.config.js file.

package.json
Parimala032: @sreenara Updated

Shreyas281299 (approved):

package.json
```diff
@@ -32,8 +32,8 @@
   ],
   "dependencies": {
     "@webex/component-adapter-interfaces": "^1.30.5",
-    "@webex/sdk-component-adapter": "1.112.9",
-    "webex": "^2.59.3",
+    "@webex/sdk-component-adapter": "1.112.10",
+    "webex": "^2.60.2",
```
rarajes2, line 36: Lets remove the caret (`^`) ?

package.json
```diff
@@ -115,7 +116,7 @@
     "prop-types": "^15.7.2",
     "react": "18.3.1",
     "react-dom": "18.3.1",
-    "webex": "^2.59.3"
+    "webex": "^2.60.2"
```
rarajes2, line 119: Why do we have `webex` in the dependency as well as devDependency?

rarajes2 (approved):
Minor comments. Please address before merging.

package.json
Parimala032: @rarajes2  This `webex` is there in the peer-dependency not in the devDependency

sreenara (approved):

PR #235, opened 2024-10-22
fix(webex): upgrading webex to new-version

Parimala032:
# **Completes** [SPARK-552246](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-552246)
## **This pull request addresses**
Upgrading the webex to newer-version which is 2.x version and the sdk-component-adapter to the newer-version as well which is 1.112.x version

- This Upgrading `webex` to newer-version is because of the fix that was made in [SDK](https://github.com/webex/webex-js-sdk/pull/3902) that needs to be in widgets 

- So the fix is about the user trying to `joinmeeting` by entering the name according to their wish but unfortunately, the user entered name was not reflected on the remote side of the meeting rather it printed the system default name so SDK fix will give the solution in the widgets that is why the `webex` is needed to upgrade in the widgets to reflect the changes that made in SDK

- Also upgrading the `sdk-component-adapter` to new version in widgets because whatever the fix does on the SDK side that version has upgraded in the `sdk-component-adapter` so the newer-version of `sdk-component-adapter` is upgrading in the `widgets` as well
## **The following scenarios were tested**

- After upgrading the version of `webex` and `sdk-component-adapter` in widgets I tested this to see whether the fix has been successful or not and it got success please do see the screen recording for more clarity 

https://github.com/user-attachments/assets/35aaa2a6-87cf-43e8-a613-c05fa2593cd6

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "@webex/component-adapter-interfaces": "^1.30.5",
         "@webex/components": "1.275.2",
         "@webex/sdk-component-adapter": "1.112.11",
-        "webex": "2.60.2"
+        "webex": "2.60.4"
       },
       "devDependencies": {
         "@babel/cli": "^7.8.4",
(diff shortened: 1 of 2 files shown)


Kesari3008:
@Parimala032 Please add test details and screen recordings or screenshots in the PR decsription

Kesari3008 (approved):

adhmenon (approved):

PR #341, opened 2024-11-22
fix(cc-store): receive webex on init, add store types

mkesavan13:
# Closes [SPARK-580566](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-580566)

# This PR addresses:

1. Revamped the Store init method with `webex` object input as an option
2. Types for Store
3. Fully covered tests for Store
4. Updating the Sample apps to receive access token and init the widgets based on that

# Testing done

1. Tested sample app with new `Store.init` parameters for webexConfig and access token on React Samples
2. Tested sample app with new `Store.init` parameters for webexConfig and access token on Web Component Samples

# Pending testing

1. `Store.init` parameters for the webex object on React Samples
1. `Store.init` parameters for the webex object on Web Component Samples

# Demo
https://app.vidcast.io/share/a394ecfa-02b3-4ab4-a2f8-f14b1afbcfe3

## Release Notes

- **New Features**
  - Introduced a user-friendly input field for entering an access token and a button to initialize SDK widgets.
  - Enhanced SDK initialization process to be dynamic based on user input.

- **Bug Fixes**
  - Improved error handling during SDK registration and initialization.

- **Tests**
  - Added a suite of unit tests for the store module, covering initialization and registration logic.

- **Chores**
  - Updated package dependencies and versioning for improved stability.


packages/contact-center/store/src/store.ts
```diff
@@ -1,8 +1,16 @@
 import {makeAutoObservable, observable} from 'mobx';
 import Webex from 'webex';
-import {IContactCenter, IAgentProfile, Team} from '@webex/plugin-cc';
+import {
+  AgentLogin,
+  IContactCenter,
+  IAgentProfile,
```
adhmenon, line 6: We might need to update this as with the latest sdk changes in Ravi's PR, we use `Profile` instead of `IAgentProfile`.

packages/contact-center/store/src/store.ts
```diff
@@ -11,20 +19,31 @@ class Store {
     makeAutoObservable(this, {cc: observable.ref});
   }
 
-  init(webexConfig: any, access_token: string): Promise<void> { 
+  registerCC(webex: WithWebex['webex']): Promise<IAgentProfile> {
```
adhmenon, line 22: I believe this will also change - only the type. Objects inside of the new `Profile` are pretty much the same, with a few additions.

Kesari3008:
@coderabbitai review

packages/contact-center/store/tests/store.test.ts
Kesari3008, line 1: We can just name the file store.ts since it's already inside tests folder

packages/contact-center/store/src/store.ts
```diff
@@ -11,20 +19,31 @@ class Store {
...
+      this.teams = response.teams;
+      this.loginOptions = response.loginVoiceOptions;
+    }).catch((error) => {
+      console.error('Error registering contact center', error);
+    });
+  }
+
+  init(options: InitParams): Promise<void> {
+    if('webex' in options) {
+      return this.registerCC(options.webex);
```
Shreyas281299, line 35: So we are assuming that ext devs will call this init only after they have received the 'ready' event. Just emphasizing so that we remember to document this. 

Lets add a comment here so that its clear for anyone who looks at the code.

packages/contact-center/store/src/store.ts
```diff
@@ -11,20 +19,31 @@ class Store {
     makeAutoObservable(this, {cc: observable.ref});
   }
 
-  init(webexConfig: any, access_token: string): Promise<void> { 
+  registerCC(webex: WithWebex['webex']): Promise<Profile> {
+    this.cc = webex.cc;
+    console.trace(this.cc.register);
```
Kesari3008, line 24: Is this console logs added intentionally or for troubleshooting purpose?

adhmenon (approved):
LGTM. Thanks for including the latest SDK changes!

packages/contact-center/store/src/store.ts
```diff
@@ -11,20 +19,31 @@ class Store {
...
+          access_token: options.access_token
         }
       });
   
       webex.once('ready', () => {
-        this.cc = webex.cc;
-        this.cc.register().then((response: IAgentProfile) => {
-          this.teams = response.teams;
-          this.loginOptions = response.loginVoiceOptions;
+        this.registerCC(webex).then(() => {
```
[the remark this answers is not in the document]
Shreyas281299: https://github.com/webex/webex-js-sdk/blob/ed1798d6f99832f91764c638d30a7e6ab36ac2da/packages/%40webex/internal-plugin-dss/src/dss.ts#L181
This is how something like this is achieved in internal-plugin-dss.

packages/contact-center/store/src/store.types.ts
```diff
@@ -0,0 +1,31 @@
...
+    registerCC(webex: WithWebex['webex']): Promise<Profile>;
+    init(params: InitParams): Promise<void>;
+}
+
+export type {
+    IContactCenter,
+    Profile,
+    Team,
+    AgentLogin,
+    WithWebex,
```
Shreyas281299, line 28: We dont need to export this type, its already exported with initParams, right?

packages/contact-center/station-login/src/station-login/index.tsx
```diff
@@ -19,7 +19,12 @@ const StationLogin: React.FunctionComponent<StationLoginProps> = observer(({onLo
   return <StationLoginPresentational {...props} />;
 });
 
-const WebStationLogin = r2wc(StationLogin);
+const WebStationLogin = r2wc(StationLogin, {
+  props: {
+    onLogin: "function",
+    onLogout: "function"
+  }
+});
```
Shreyas281299, line 27: Was this not caught because we didn't build web-components-samples and test in the web-component-samples? This is a breaking change because react and web-component under the same package have different behaviour. 

To catch these types of errors should we have a `Build Samples` step in the pipeline? Since onLogin and onLogout are mandatory arguments, the build should have failed. Il check and update it here.

Shreyas281299: We should also see how we can test this in unit testing. Maybe an argument check AFTER converting the component into a web component

packages/contact-center/store/src/store.ts
```diff
@@ -11,20 +19,31 @@ class Store {
...
 
-  init(webexConfig: any, access_token: string): Promise<void> { 
+  registerCC(webex: WithWebex['webex']): Promise<Profile> {
+    this.cc = webex.cc;
+    console.trace(this.cc.register);
+    return this.cc.register().then((response: Profile) => {
+      this.teams = response.teams;
+      this.loginOptions = response.loginVoiceOptions;
+    }).catch((error) => {
+      console.error('Error registering contact center', error);
```
Kesari3008, line 29: Won't we need to reject the promise in error case? Also will return take of resolving the promise.

packages/contact-center/store/src/store.types.ts
```diff
@@ -0,0 +1,31 @@
+import {AgentLogin, IContactCenter, Profile, Team} from '@webex/plugin-cc';
+
+interface WithWebex {
+    webex: { cc: IContactCenter };
+}
+
+interface WithWebexConfigAndToken {
```
Kesari3008, line 7: We can simplify the name of this interface: WebexConfig or StoreConfig should be ok

docs/react-samples/src/App.tsx
```diff
@@ -1,26 +1,20 @@
...
+  const [isSdkReady, setIsSdkReady] = useState(false);
+  const [accessToken, setAccessToken] = useState("");
 
   const webexConfig = {
       fedramp: false,
       logger: {
-      level: 'log'  // TODO: We will add more logging levels later and set the righ levels
+        level: 'log'
       },
+      credentials: undefined,
```
Kesari3008, line 15: Why do we need to add credentials here ?

docs/react-samples/src/App.tsx
```diff
@@ -32,8 +26,29 @@ function App() {
...
+      >Init Widgets</button>
+      {/* write code to check if sdk is ready and load components */}
+      {
+        isSdkReady && (
+          <>
+            <StationLogin  onLogin={onLogin} onLogout={onLogout} />
+            <UserState />
+          </>
+        )
+      }
```
Shreyas281299, line 51: I don't think this is possible. In the samples app, we cant pass our custom Webex object so we can remove this. We should definitely test this. I'm not sure how we will keep this in check in future.

Shreyas281299: We can remove the isSDKReady state then.

docs/web-component-samples/index.html
```diff
@@ -4,23 +4,61 @@
...
   </body>
   <script>
-    const token = '';
-    
-    const webexConfig = {
-        fedramp: false,
-        logger: {
-        level: 'log'  // TODO: We will add more logging levels later and set the righ levels
-        },
+    const widgetsContainer = document.getElementById('widgets-container');
```
Kesari3008, line 29: This file will keep increasing so does it make sense to create separate .js file and have that added in this script section

packages/contact-center/store/tests/store.test.ts
```diff
@@ -0,0 +1,104 @@
...
+  });
+
+  describe('init', () => {
+    it('should call registerCC if webex is in options', async () => {
+      const initParams = { webex: mockWebex };
+      jest.spyOn(store, 'registerCC').mockResolvedValue();
+
+      await store.init(initParams);
+
+      expect(store.registerCC).toHaveBeenCalledWith(mockWebex);
```
Shreyas281299, line 74: Can we add a check to ensure that it doesn't call, Webex.init in this case

Shreyas281299 (changes requested):
Some comments and some thoughts. Please have a look.

Kesari3008 (changes requested):
Please take a look at coderabbit comments as well

packages/contact-center/store/src/store.types.ts
Shreyas281299: This can be an internal type as the type that is exported is initParams. But im okay with making the name simpler also.

docs/web-component-samples/index.html
Shreyas281299: We have an index.js file that is importing the widgets, maybe we can move this code in that file.

packages/contact-center/store/src/store.types.ts
```diff
@@ -0,0 +1,31 @@
...
+    webexConfig: any; // Replace 'any' with the actual type of webexConfig
+    access_token: string;
+}
```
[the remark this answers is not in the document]
mkesavan13: Do not want to setup a config type as this is only needed config for our samples whereas the users might come up with different config. Will do later

packages/contact-center/store/package.json
```diff
@@ -18,12 +18,11 @@
     "test:unit": "jest"
   },
   "dependencies": {
-    "@webex/plugin-cc": "3.5.0-wxcc.2",
     "mobx": "6.13.5",
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "typescript": "5.6.3",
-    "webex": "3.6.0-wxcc.3"
+    "webex": "3.6.0-wxcc.5"
```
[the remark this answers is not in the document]
mkesavan13: `@webex/plugin-cc` is a dependency of `webex` package already and the imports will still work alright. It was a redundant dependency. That is why removed it.

docs/react-samples/src/App.tsx
```diff
@@ -32,8 +26,29 @@ function App() {
...
     <>
       <h1>Contact Center widgets in a react app</h1>
-      <StationLogin  onLogin={onLogin} onLogout={onLogout} />
-      <UserState />
+      <input
+        type="text"
+        placeholder="Enter your access token"
+        value={accessToken}
+        onChange={(e) => setAccessToken(e.target.value)}
+      />
```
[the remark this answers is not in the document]
mkesavan13: This is a sample app change. Neither validation, nor aria-label is required. So going to ignore this.

docs/react-samples/src/App.tsx
```diff
@@ -32,8 +26,29 @@ function App() {
...
+      <button
+        disabled={accessToken.trim() === ""}
+        onClick={() => {
+          store.init({webexConfig, access_token: accessToken}).then(() => {
+            setIsSdkReady(true);
+          });
+        }}
+      >Init Widgets</button>
```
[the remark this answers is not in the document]
mkesavan13: This is a sample app change. Not going to do it. This will make it obligated to do the same thing on Web Component Samples. Will become overhead and an overkill for a sample app

docs/web-component-samples/index.html
```diff
@@ -4,23 +4,61 @@
...
   </head>
   <body>
     <h1>Contact Center widgets as web-component</h1>
+    <input
+      type="text"
+      placeholder="Enter your access token"
+      id="access_token_elem"
+      onkeyup="switchButtonState()"
+    />
+    <button onclick="initWidgets()" disabled>Init Widgets</button>
```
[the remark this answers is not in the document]
mkesavan13: Again an overkill for sample app. Still, since it is a small suggestion, made the changes.

docs/web-component-samples/index.html
```diff
@@ -4,23 +4,61 @@
...
-        level: 'log'  // TODO: We will add more logging levels later and set the righ levels
-        },
+    const widgetsContainer = document.getElementById('widgets-container');
+    const accessTokenElem = document.getElementById('access_token_elem');
+    const ccStationLogin = document.getElementById('cc-station-login');
+
+    function switchButtonState(){
+      const buttonElem = document.querySelector('button');
+      buttonElem.disabled = accessTokenElem.value.trim() === '';
+    }
```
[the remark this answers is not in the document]
mkesavan13: Overkill for sample app. Not doing it

docs/web-component-samples/index.html
```diff
@@ -4,23 +4,61 @@
...
-    const token = '';
-    
-    const webexConfig = {
-        fedramp: false,
-        logger: {
-        level: 'log'  // TODO: We will add more logging levels later and set the righ levels
-        },
+    const widgetsContainer = document.getElementById('widgets-container');
+    const accessTokenElem = document.getElementById('access_token_elem');
+    const ccStationLogin = document.getElementById('cc-station-login');
```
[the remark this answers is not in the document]
mkesavan13: Overkill for a sample app. Not doing this

packages/contact-center/store/tests/store.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+      await store.init(initParams);
+
+      expect(Webex.init).toHaveBeenCalledWith({
+        config: initParams.webexConfig,
+        credentials: { access_token: initParams.access_token }
+      });
+      expect(store.registerCC).toHaveBeenCalledWith(expect.any(Object));
+    });
+
```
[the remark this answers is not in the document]
mkesavan13: Not required. Typescript handles this

packages/contact-center/store/src/store.ts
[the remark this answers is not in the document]
mkesavan13: We have tested this working alright practically. However, I agree that we need that guard. Will add it

packages/contact-center/store/src/store.types.ts
mkesavan13: In store.ts, we have registerCC method which receives `webex` object. Although it is typed as `any` for now, if we update its type later, wanted it to work automatically. So referred the type from `WithWebex` like this:

```javascript
registerCC(webex: WithWebex['webex']): Promise<void> {
    //register code
}
```

packages/contact-center/store/src/store.types.ts
```diff
@@ -0,0 +1,31 @@
+import {AgentLogin, IContactCenter, Profile, Team} from '@webex/plugin-cc';
+
+interface WithWebex {
+    webex: { cc: IContactCenter };
+}
+
+interface WithWebexConfigAndToken {
+    webexConfig: any; // Replace 'any' with the actual type of webexConfig
```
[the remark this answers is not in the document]
mkesavan13: We will have to do it later by obtaining the types from SDK. This will be a later TODO

packages/contact-center/store/src/store.types.ts
```diff
@@ -0,0 +1,31 @@
...
+    access_token: string;
+}
+  
+type InitParams = WithWebex | WithWebexConfigAndToken;
+
+interface IStore {
+    teams: Team[];
+    loginOptions: string[];
```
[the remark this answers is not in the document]
mkesavan13: I'd be more happy to get this from the SDK rather than typing here as it would be subject to change. Leaving it as it is for now

CC - @Kesari3008 @rsarika

packages/contact-center/store/tests/store.test.ts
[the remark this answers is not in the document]
mkesavan13: Done. The problem here was that the jest config in `package.json` did not have `testMatch`

Therefore, I assumed we'd go with the default way of naming them. Updated now

packages/contact-center/station-login/src/station-login/index.tsx
mkesavan13: This was not caught because the Web Component samples weren't tested.

This should be the ownership of both reviewers and authors to ensure that WC Samples are tested. The way to catch this would be to introduce UI testing like Cypress

@Shreyas281299

docs/react-samples/src/App.tsx
mkesavan13: Could you please mention what is not possible?

If you look at the button onclick, you'll understand what's going on here. The `isSdkReady` is a state that's set after the `store.init` promise resolves. This is just to ensure the components load only after the store is initialised

packages/contact-center/store/tests/store.test.ts
mkesavan13: Done

packages/contact-center/store/src/store.types.ts
mkesavan13: It is internal and kept it so just to make the InitParams type to be descriptive:

```
type InitParams = WithWebex | WithWebexConfigAndToken;
```

Changed it to `WithWebexConfig` considering the token as part of the config anyway

docs/react-samples/src/App.tsx
mkesavan13: I added it for testing the credentials config and forgot to remove it. Good catch. Thank you.

docs/web-component-samples/index.html
mkesavan13: Done.

packages/contact-center/store/src/store.types.ts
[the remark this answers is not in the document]
Kesari3008: This array will change based on what is received from the agent config so it makes sense to keep it as a string array

packages/contact-center/store/src/store.ts
```diff
@@ -11,20 +19,31 @@ class Store {
     makeAutoObservable(this, {cc: observable.ref});
   }
 
-  init(webexConfig: any, access_token: string): Promise<void> { 
+  registerCC(webex: WithWebex['webex']): Promise<Profile> {
+    this.cc = webex.cc;
+    console.trace(this.cc.register);
+    return this.cc.register().then((response: Profile) => {
```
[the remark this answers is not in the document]
Kesari3008: This is not addressed, do we not need to fix this

Kesari3008 (approved):
Looks good to me. But it would be good to look at the code rabbit comments and address them wherever applicable

docs/web-component-samples/index.html
```diff
@@ -4,23 +4,61 @@
...
+      }
+
+      store.init({
+        webexConfig,
+        access_token: accessTokenElem.value.trim()
+      }).then(() => {
+        ccStationLogin.onLogin = loginSuccess;
+        ccStationLogin.onLogout = logoutSuccess;
+        widgetsContainer.classList.remove('disabled');
+      });
```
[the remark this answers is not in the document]
Kesari3008: This should be taken care of even if it's sample app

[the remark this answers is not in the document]
mkesavan13: This is addressed here: https://github.com/webex/widgets/pull/341/files#diff-456a86c88aca834f63f52ef26e3311ebab04a063aa14db609cb74e88558960faR25

@Kesari3008

docs/react-samples/src/App.tsx
Shreyas281299: My bad. I got confused. I thought in the samples we are getting the webex object from devs. Thats what I thought is not possible. This is the flow where we get the access token. Resolving this.

Shreyas281299 (approved):

PR #356, opened 2025-01-10
fix: cleanup and using mobx-react-lite

rarajes2:
JIRA - [SPARK-601658](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-601658)

- Fixes the build and other commands in the `cc-task` package
- Fixes the dependencies in multiple packages to ensure proper dependencies
- Uses `mobx-react-lite` instead of `mobx-react` as we are only using functional components
- Fixes broken UT with jest config

## Summary by CodeRabbit

- **Dependency Updates**
  - Replaced `mobx-react` with `mobx-react-lite` in multiple packages
  - Updated React-related peer dependencies

- **Build Process Improvements**
  - Added `clean:dist` step before webpack builds across multiple packages
  - Enhanced build scripts to ensure clean build environments

- **Testing Configuration**
  - Updated Jest configurations to handle CSS, LESS, and SCSS files

- **Webpack Configuration**
  - Added external dependencies configuration to reduce bundle size

### DIFF
--- a/packages/contact-center/cc-widgets/package.json
+++ b/packages/contact-center/cc-widgets/package.json
@@ -18,7 +18,7 @@
     "clean": "rm -rf dist && rm -rf node_modules",
     "clean:dist": "rm -rf dist",
     "build": "yarn run -T tsc",
-    "build:src": "webpack",
+    "build:src": "yarn run clean:dist && webpack",
     "build:watch": "webpack --watch",
     "test:unit": "jest"
   },
@@ -45,7 +45,6 @@
     "file-loader": "6.2.0",
     "jest": "29.7.0",
     "jest-environment-jsdom": "29.7.0",
-    "mobx-react": "9.1.1",
     "os-browserify": "^0.3.0",
     "process": "^0.11.10",
     "querystring-es3": "^0.2.1",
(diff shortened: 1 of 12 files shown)


Shreyas281299:
@coderabbit review

packages/contact-center/cc-widgets/package.json
```diff
@@ -18,7 +18,7 @@
     "clean": "rm -rf dist && rm -rf node_modules",
     "clean:dist": "rm -rf dist",
     "build": "yarn run -T tsc",
```
Shreyas281299, line 20: Are we still using this build command?

Shreyas281299 (approved):
Looks good. Just a question - What was causing the test failures??

packages/contact-center/cc-widgets/package.json
rarajes2: Not using but keeping it if we need to test something. If we don't use it for quite some time, we can remove this.

rarajes2:
> Looks good. Just a question - What was causing the test failures??

We needed to update the jest config

```
...
 "moduleNameMapper": {
     "^.+\\.(css|less|scss)$": "babel-jest"
  }
```

PR #378, opened 2025-02-05
fix(cc-components): replace momentum-ui with momentum-design lib

mkesavan13:
# Closes Ad-hoc

Replaces old `@momentum-ui/web-components` with new `@momentum-design/components` repo

# Screen Recording of WC Samples

https://github.com/user-attachments/assets/1e8fca5d-7fb6-4225-b3c5-478863af0daf


packages/contact-center/user-state/tests/user-state/index.tsx
```diff
@@ -27,6 +25,10 @@ describe('UserState Component', () => {
     };
     global.Worker = jest.fn(() => workerMock);
     global.URL.createObjectURL = jest.fn(() => 'blob:http://localhost:3000/12345');
+
+    if (typeof window.HTMLElement.prototype.attachInternals !== 'function') {
+      window.HTMLElement.prototype.attachInternals = jest.fn() as any;
+    }
```
mkesavan13, line 31: This is similar to how we mocked the web components earlier. We do not need to test any part of the UI lib here as we do that in the `cc-components` tests anyway

packages/contact-center/cc-components/src/components/UserState/user-state.tsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 
 import {IUserState} from './user-state.types';
 import {formatTime} from '../../utils'
+import { Button } from '@momentum-design/components/dist/react';
```
Shreyas281299, line 5: Why are they not exporting the component?? This way of using isn't the best choice

packages/contact-center/cc-components/src/wc.ts
```diff
@@ -1,8 +1,10 @@
 import r2wc from '@r2wc/react-to-web-component';
 import UserStateComponent from './components/UserState/user-state';
-import '@momentum-ui/web-components';
-import '@momentum-ui/icons/css/momentum-ui-icons.min.css';
-import '@momentum-ui/core/css/momentum-ui.min.css';
+import '@momentum-design/components/components/button';
+import '@momentum-design/components/components/themeprovider';
+import '@momentum-design/components/components/iconprovider';
+import '@momentum-design/fonts/dist/css/fonts.css';
+import '@momentum-design/tokens/dist/css/components/complete.css';
```
pagour98, line 7: @mkesavan13 I think we can import the necessary css libraries from @momentum-design that we really need, if we import the complete.css this may increase the bundle size.

Reference:- https://momentum-design.github.io/momentum-design/en/components/
Screenshot 2025-02-06 at 14 51 58

pagour98 (changes requested):

packages/contact-center/cc-components/src/components/UserState/user-state.tsx
Shreyas281299: We are still fixing

packages/contact-center/cc-components/src/wc.ts
mkesavan13: @pagour98 - Right now, we are looking to import all of them anyway. There's no component specific css. All the components come as part of complete.css. We need Fonts, Typography, Effects, Dark and Light themes anyway.

packages/contact-center/cc-components/package.json
```diff
@@ -20,7 +20,7 @@
     "build": "yarn run -T tsc",
     "build:src": "yarn run clean:dist && webpack",
     "build:watch": "webpack --watch",
-    "test:unit": "jest"
+    "test:unit:broken": "jest"
```
mkesavan13, line 23: We've marked it broken temporarily. We are working with Christoph's team to fix it and there's a Bug Kanban board sticky for it.

pagour98 (approved):

PR #389, opened 2025-02-18
feat(widgets): added-agent-rona-popup

adhmenon:
## ISSUE
* In widgets, we were receiving the agent rona related event but we were not handling it in the sample app.

## FIX
* Code has been added to include a `taskRejectionCallback` handler inside of the widgets store so users can specify their own custom task rejection method
* For our use case, we show a popup whihch allows users to set a state inside of our sample app (this is widgets sample app specific change)

## Vidcasts
* [Working with WebRTC in web components](https://app.vidcast.io/share/d451f694-43db-407a-9246-5218769f50c3)
* [Working with WebRTC in react components](https://app.vidcast.io/share/cd9e130e-d1fd-4689-bf1a-d733817975e3)
* [Working in EXTERNAL config](https://app.vidcast.io/share/e5a4d7d7-be0f-4018-85f7-f635d8200ce3)

### Manual Test Cases
* Tested with webRTC in both web and react components
* Tested with external type in both web and react
* Tested by keeping both web and react sample pages open and used external type (we got popup in both as expected)
* Tested that the user state is not getting messed up with this (cross verified with agent desktop)

Test Doc Link - https://confluence-eng-gpk2.cisco.com/conf/display/WSDK/CC+Widgets+Test+Plan

### DIFF
--- a/packages/contact-center/store/src/store.types.ts
+++ b/packages/contact-center/store/src/store.types.ts
@@ -52,7 +52,6 @@ interface IStore {
 
 interface IStoreWrapper extends IStore{
     store: IStore;
-    onTaskRejected?: (reason: string) => void;
     setCurrentTask(task: any): void;
     setWrapupRequired(value: boolean): void;
     setTaskList(taskList: ITask[]): void;
--- a/packages/contact-center/store/src/store.types.ts
+++ b/packages/contact-center/store/src/store.types.ts
@@ -52,6 +52,7 @@ interface IStore {
 
 interface IStoreWrapper extends IStore{
     store: IStore;
+    onTaskRejected?: (reason: string) => void;
     setCurrentTask(task: any): void;
     setWrapupRequired(value: boolean): void;
     setTaskList(taskList: ITask[]): void;
--- a/packages/contact-center/store/src/storeEventsWrapper.ts
+++ b/packages/contact-center/store/src/storeEventsWrapper.ts
@@ -5,6 +5,7 @@ import {runInAction} from 'mobx';
 
 class StoreWrapper implements IStoreWrapper {
   store: IStore;
+  onTaskRejected?: (reason: string) => void;
 
   constructor() {
     this.store = Store.getInstance();
@@ -183,7 +184,13 @@ class StoreWrapper implements IStoreWrapper {
 
     // When we receive TASK_REJECT sdk changes the agent status
     // When we receive TASK_REJECT that means the task was not accepted by the agent and we wont need wrap up
-    task.on(TASK_EVENTS.TASK_REJECT, () => this.handleTaskRemove(task.data.interactionId));
+    task.on(TASK_EVENTS.TASK_REJECT, (reason: string) => {
+      // Call the user defined callback first if available
+      if (this.onTaskRejected) {
+        this.onTaskRejected(reason || 'No reason provided');
+      }
+      this.handleTaskRemove(task.data.interactionId);
+    });
 
     this.setTaskList([...this.store.taskList, task]);
   };
@@ -214,7 +221,13 @@ class StoreWrapper implements IStoreWrapper {
 
     // When we receive TASK_REJECT sdk changes the agent status
     // When we receive TASK_REJECT that means the task was not accepted by the agent and we wont need wrap up
-    task.on(TASK_EVENTS.TASK_REJECT, () => this.handleTaskRemove(task.data.interactionId));
+    task.on(TASK_EVENTS.TASK_REJECT, (reason: string) => {
+      // Call the user defined callback first if available
+      if (this.onTaskRejected) {
+        this.onTaskRejected(reason || 'No reason provided');
+      }
+      this.handleTaskRemove(task.data.interactionId);
+    });
 
     if (!this.store.taskList.some((t) => t.data.interactionId === task.data.interactionId)) {
       this.setTaskList([...this.store.taskList, task]);
--- a/packages/contact-center/store/src/storeEventsWrapper.ts
+++ b/packages/contact-center/store/src/storeEventsWrapper.ts
@@ -115,6 +115,10 @@ class StoreWrapper implements IStoreWrapper {
     this.store.wrapupCodes = wrapupCodes;
   };
 
+  setTaskRejected = (callback: ((reason: string) => void) | undefined): void => {
+    this.onTaskRejected = callback;
+  };
+
   init(options: InitParams): Promise<void> {
     return this.store.init(options, this.setupIncomingTaskHandler);
   }
--- a/packages/contact-center/store/src/storeEventsWrapper.ts
+++ b/packages/contact-center/store/src/storeEventsWrapper.ts
@@ -184,13 +184,9 @@ class StoreWrapper implements IStoreWrapper {
 
     // When we receive TASK_REJECT sdk changes the agent status
     // When we receive TASK_REJECT that means the task was not accepted by the agent and we wont need wrap up
-    task.on(TASK_EVENTS.TASK_REJECT, (reason: string) => {
-      // Call the user defined callback first if available
-      if (this.onTaskRejected) {
-        this.onTaskRejected(reason || 'No reason provided');
-      }
-      this.handleTaskRemove(task.data.interactionId);
-    });
+    task.on(TASK_EVENTS.TASK_REJECT, (reason: string) => 
+      this.handleTaskReject(task, reason)
+    );
 
     this.setTaskList([...this.store.taskList, task]);
   };
@@ -221,13 +217,9 @@ class StoreWrapper implements IStoreWrapper {
 
     // When we receive TASK_REJECT sdk changes the agent status
     // When we receive TASK_REJECT that means the task was not accepted by the agent and we wont need wrap up
-    task.on(TASK_EVENTS.TASK_REJECT, (reason: string) => {
-      // Call the user defined callback first if available
-      if (this.onTaskRejected) {
-        this.onTaskRejected(reason || 'No reason provided');
-      }
-      this.handleTaskRemove(task.data.interactionId);
-    });
+    task.on(TASK_EVENTS.TASK_REJECT, (reason: string) => 
+      this.handleTaskReject(task, reason)
+    );
 
     if (!this.store.taskList.some((t) => t.data.interactionId === task.data.interactionId)) {
       this.setTaskList([...this.store.taskList, task]);
@@ -247,6 +239,13 @@ class StoreWrapper implements IStoreWrapper {
       return;
     }
   };
+
+  handleTaskReject = (task: ITask, reason: string) => {
+    if (this.onTaskRejected) {
+      this.onTaskRejected(reason || 'No reason provided');
+    }
+    this.handleTaskRemove(task.data.interactionId);
+  }
 
   setupIncomingTaskHandler = (ccSDK: any) => {
     ccSDK.on(TASK_EVENTS.TASK_INCOMING, this.handleIncomingTask);
--- a/packages/contact-center/store/src/storeEventsWrapper.ts
+++ b/packages/contact-center/store/src/storeEventsWrapper.ts
@@ -189,7 +189,7 @@ class StoreWrapper implements IStoreWrapper {
     // When we receive TASK_REJECT sdk changes the agent status
     // When we receive TASK_REJECT that means the task was not accepted by the agent and we wont need wrap up
     task.on(TASK_EVENTS.TASK_REJECT, (reason: string) => 
-      this.handleTaskReject(task, reason)
+      this.handleTaskReject(task.data.interactionId, reason)
     );
 
     this.setTaskList([...this.store.taskList, task]);
@@ -222,7 +222,7 @@ class StoreWrapper implements IStoreWrapper {
     // When we receive TASK_REJECT sdk changes the agent status
     // When we receive TASK_REJECT that means the task was not accepted by the agent and we wont need wrap up
     task.on(TASK_EVENTS.TASK_REJECT, (reason: string) => 
-      this.handleTaskReject(task, reason)
+      this.handleTaskReject(task.data.interactionId, reason)
     );
 
     if (!this.store.taskList.some((t) => t.data.interactionId === task.data.interactionId)) {
@@ -244,11 +244,11 @@ class StoreWrapper implements IStoreWrapper {
     }
   };
 
-  handleTaskReject = (task: ITask, reason: string) => {
+  handleTaskReject = (taskId: string, reason: string) => {
     if (this.onTaskRejected) {
       this.onTaskRejected(reason || 'No reason provided');
     }
-    this.handleTaskRemove(task.data.interactionId);
+    this.handleTaskRemove(taskId);
   }
 
   setupIncomingTaskHandler = (ccSDK: any) => {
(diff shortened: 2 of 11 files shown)


packages/contact-center/store/src/store.types.ts
```diff
@@ -52,6 +52,7 @@ interface IStore {
 
 interface IStoreWrapper extends IStore{
     store: IStore;
+    onTaskRejected?: (reason: string) => void;
```
adhmenon, line 55: As discussed, here I am exposing a callback so users can specify their own `task rejected` methods.

packages/contact-center/store/src/storeEventsWrapper.ts
```diff
@@ -5,6 +5,7 @@ import {runInAction} from 'mobx';
 
 class StoreWrapper implements IStoreWrapper {
   store: IStore;
+  onTaskRejected?: (reason: string) => void;
```
adhmenon, line 8: I kept it as void return type and have added this in the wrapper as well.
I felt we should input it here as store will be used regardless of whether the user is using task list or incoming task. Further, we are handling the event here so I added this here.
Open to any suggestion however.

packages/contact-center/store/src/storeEventsWrapper.ts
```diff
@@ -214,7 +221,13 @@ class StoreWrapper implements IStoreWrapper {
 
     // When we receive TASK_REJECT sdk changes the agent status
     // When we receive TASK_REJECT that means the task was not accepted by the agent and we wont need wrap up
-    task.on(TASK_EVENTS.TASK_REJECT, () => this.handleTaskRemove(task.data.interactionId));
+    task.on(TASK_EVENTS.TASK_REJECT, (reason: string) => {
```
adhmenon, line 224: As explained above, we are checking if task rejected is there (optional again), but we are cleaning up the task either ways.

packages/contact-center/store/tests/storeEventsWrapper.ts
```diff
@@ -608,5 +608,28 @@ describe('storeEventsWrapper', () => {
       expect(storeWrapper['cc'].off).toHaveBeenCalledWith(CC_EVENTS.AGENT_STATE_CHANGE, expect.any(Function));
       expect(storeWrapper['cc'].off).toHaveBeenCalledWith(CC_EVENTS.AGENT_MULTI_LOGIN, expect.any(Function));
     });
+
+    it('should handle task rejection event and call onTaskRejected with the provided reason', () => {
```
adhmenon, line 612: Test to check if the reason / rejected task etc. are coming properly.

widgets-samples/cc/samples-cc-react-app/src/App.tsx
```diff
@@ -78,6 +81,54 @@ function App() {
...
+
+  const handlePopoverSubmit = () => {
+    if (selectedState) {
+      changeAgentState(selectedState);
+    }
+    setShowRejectedPopup(false);
+    setSelectedState('');
+  };
+
+  useEffect(() => {
```
adhmenon, line 121: Wanted to add this code here as I wanted to set up the callback on the first mounting itself.

widgets-samples/cc/samples-cc-react-app/src/App.tsx
```diff
@@ -78,6 +81,54 @@ function App() {
     setSelectedWidgets((prev) => ({...prev, [name]: checked}));
   };
 
+  const changeAgentState = (newState: string) => {
+    // If the selected state is "Available", we pass it as is; otherwise, we consider it "Idle".
+    const chosenState = newState === 'Available' ? 'Available' : 'Idle';
+    // In the idle codes, we need to search for the 'Idle' state with code name 'Meeting'.
+    const lookupCodeName = newState === 'Available' ? 'Available' : 'Meeting';
```
adhmenon, line 88: Sadly the above 2 are needed as backend only recognises `Available` and `Idle`.
Also for idle codes, we sadly only have the above 2, hence we have to add this check (note that this is only for sample app, however). We are doing the same in the cc sdk sample app.

widgets-samples/cc/samples-cc-react-app/src/App.tsx
```diff
@@ -189,9 +240,34 @@ function App() {
                 </>
               )}
             </>
+          )}  
+
+        {showRejectedPopup && (
```
adhmenon, line 245: We simply show the popup here also.

widgets-samples/cc/samples-cc-wc-app/app.js
```diff
@@ -126,13 +131,71 @@ function onTaskDeclined() {
...
+      store.setLastStateChangeTimestamp(new Date(response.data.lastStateChangeTimestamp));
+      console.log('Agent state updated to', chosenState);
+    })
+    .catch(function(error) {
+      console.error('Error updating agent state:', error);
+    });
+}
+
+// Helper to show the task rejected popup.
+function showTaskRejectedPopup(reason) {
```
adhmenon, line 177: Again, this is so that we can get the popup similar to react sample app, open to suggestions here also.

widgets-samples/cc/samples-cc-wc-app/app.js
```diff
@@ -27,6 +28,10 @@ themeElem.addEventListener('change', () => {
   );
 });
 
+store.onTaskRejected = function(reason) {
```
adhmenon, line 31: Add the callback here itself, felt it would be best.

adhmenon:
@coderabbitai review

adhmenon:
@coderabbitai resolved all your comments.

packages/contact-center/store/src/store.types.ts
Shreyas281299: In this file below you will see the TASK_EVENTS, lets use them and remove the taskEvents from @webex/cc-tasks

packages/contact-center/task/src/task.types.ts
```diff
@@ -120,6 +120,7 @@ export enum TASK_EVENTS {
   TASK_RESUME = 'task:resume',
   TASK_END = 'task:end',
   TASK_WRAPUP = 'task:wrapup',
+  TASK_REJECT = 'task:rejected',
```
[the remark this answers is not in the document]
Shreyas281299: I have missed to remove these from here in my storeWrapper PR

widgets-samples/cc/samples-cc-react-app/src/App.tsx
```diff
@@ -78,6 +81,54 @@ function App() {
...
+    }
+    setShowRejectedPopup(false);
+    setSelectedState('');
+  };
+
+  useEffect(() => {
+    store.onTaskRejected = (reason: string) => {
+      setRejectedReason(reason);
+      setShowRejectedPopup(true);
+    };
```
Shreyas281299, line 125:
```suggestion
    const handleTaskRejected =(reason: string) => {
      setRejectedReason(reason);
      setShowRejectedPopup(true);
    };
    
    store.setOnTaskRejected(handleTaskRejected) 
```

Lets create a setter for the callBack rather than asking users to set a state in the store.

widgets-samples/cc/samples-cc-react-app/src/App.tsx
```diff
@@ -78,6 +81,54 @@ function App() {
...
+
+    return () => {
+      store.onTaskRejected = undefined;
```
Shreyas281299, line 128:
```suggestion
      store.setOnTaskRejected(undefined)
```

widgets-samples/cc/samples-cc-wc-app/app.js
```diff
@@ -136,3 +141,57 @@ function onEnd() {
...
+        <h2>Task Rejected</h2>
+        <p>Reason: ${reason || 'No reason provided'}</p>
+        <select id="state-select">
+          <option value="">Select a state</option>
+          <option value="Available">Available</option>
+          <option value="Busy">Busy</option>
+          <option value="On Break">On Break</option>
+        </select>
+        <button id="submit-button">Submit</button>
+      </div>
```
Shreyas281299, line 185: Would it be better to move this to the html and keep this div as disabled, and on the invocation of the handleTaskRejected we can just enable it

widgets-samples/cc/samples-cc-wc-app/index.html
```diff
@@ -51,6 +61,7 @@ <h1>Contact Center widgets as web-component</h1>
         <br />
         <button onclick="initWidgets()" disabled>Init Widgets</button>
         <div id="widgets-container"></div>
+        <div id="popup-container" style="display: none;"></div>
```
Shreyas281299, line 64: Since the display is already non here why not put the whole HTML object here. 

div id="task-rejected-popup">
        <h2>Task Rejected</h2>
        <p>Reason: ${reason || 'No reason provided'}</p>
        <select id="state-select">
          <option value="">Select a state</option>
          <option value="Available">Available</option>
          <option value="Busy">Busy</option>
          <option value="On Break">On Break</option>
        </select>
        <button id="submit-button">Submit</button>
      </div>

Shreyas281299, line 64:
```suggestion
        <div id="popup-container" style="display: none;" disabled>
         <div id="task-rejected-popup">
          <h2>Task Rejected</h2>
          <p>Reason: ${reason || 'No reason provided'}</p>
          <select id="state-select">
            <option value="">Select a state</option>
            <option value="Available">Available</option>
            <option value="Busy">Busy</option>
            <option value="On Break">On Break</option>
          </select>
          <button id="submit-button">Submit</button>
         </div>
        </div>
```

Shreyas281299:
Please do a testing with multiple tab with the same user

- Have 2 tabs open with the same user logged in as EXT in both of them
- See if the RONA is MODAL is working as expected, I feel when we get the modal on 1 tab and we set the state on that window, the UI in the second window will not be in sync, the RONA modal will be still set in the second modal.

adhmenon:
> Please do a testing with multiple tab with the same user
> 
>     * Have 2 tabs open with the same user logged in as EXT in both of them
> 
>     * See if the RONA is MODAL is working as expected, I feel when we get the modal on 1 tab and we set the state on that window, the UI in the second window will not be in sync, the RONA modal will be still set in the second modal.

Yes, I have tested this scenario and as mentioned in the test plan - the state syncs, however we do see modals in both windows. I think this behaviour is alright (considering it's not widget specific) and that in each tab we essentially are handling the modal (again a sample app issue).

packages/contact-center/task/tests/CallControl/index.tsx
```diff
@@ -21,6 +21,9 @@ jest.mock('@webex/cc-store', () => ({
     end: jest.fn(() => Promise.resolve()),
     wrapup: jest.fn(() => Promise.resolve()),
   },
+  TASK_EVENTS: {
```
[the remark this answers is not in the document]
Shreyas281299: For testing its fine.

packages/contact-center/store/src/storeEventsWrapper.ts
Shreyas281299, line 8: Still curious if we need this or not? Can you try to remove this and see if something breaks? If it breaks something then we can add this ad leave it.

Shreyas281299 (approved):
Changes look good,.

Shreyas281299:
> > Please do a testing with multiple tab with the same user
> > ```
> > * Have 2 tabs open with the same user logged in as EXT in both of them
> > 
> > * See if the RONA is MODAL is working as expected, I feel when we get the modal on 1 tab and we set the state on that window, the UI in the second window will not be in sync, the RONA modal will be still set in the second modal.
> > ```
> 
> Yes, I have tested this scenario and as mentioned in the test plan - the state syncs, however we do see modals in both windows. I think this behaviour is alright (considering it's not widget specific) and that in each tab we essentially are handling the modal (again a sample app issue). Also, the end user would control the behaviour so I think from widgets side we are good. Sample app... perhaps this is something we might have to leave... not sure what else can be done otherwise ... since the popup could be opened via web components sample and react sample also.

Just a question- when we have modal in both the sample pages and we click say change my state to available, do both the modals go away? or only in 1 tab the modal goes away.

adhmenon:
> > > Please do a testing with multiple tab with the same user
> > > ```
> > > * Have 2 tabs open with the same user logged in as EXT in both of them
> > > 
> > > * See if the RONA is MODAL is working as expected, I feel when we get the modal on 1 tab and we set the state on that window, the UI in the second window will not be in sync, the RONA modal will be still set in the second modal.
> > > ```
> > 
> > 
> > Yes, I have tested this scenario and as mentioned in the test plan - the state syncs, however we do see modals in both windows. I think this behaviour is alright (considering it's not widget specific) and that in each tab we essentially are handling the modal (again a sample app issue). Also, the end user would control the behaviour so I think from widgets side we are good. Sample app... perhaps this is something we might have to leave... not sure what else can be done otherwise ... since the popup could be opened via web components sample and react sample also.
> 
> Just a question- when we have modal in both the sample pages and we click say change my state to available, do both the modals go away? or only in 1 tab the modal goes away.

Only 1 goes away actually.
I could maybe put a check in both sample apps to see listen for state change event and if any modal is open close it... but if a user opens both web and react components it might not work. Let me check.

packages/contact-center/store/src/storeEventsWrapper.ts
```diff
@@ -214,7 +221,9 @@ class StoreWrapper implements IStoreWrapper {
 
     // When we receive TASK_REJECT sdk changes the agent status
     // When we receive TASK_REJECT that means the task was not accepted by the agent and we wont need wrap up
-    task.on(TASK_EVENTS.TASK_REJECT, () => this.handleTaskRemove(task.data.interactionId));
+    task.on(TASK_EVENTS.TASK_REJECT, (reason: string) =>
```
Kesari3008, line 224: Task hydrate happens if reload must have happened in an incomplete state, just want to understand why we might receive a Task reject at this stage. If there was an incoming call before the reload and the reload happened, we need not to retain the state as that call should be cleared and since it never connected, we anyway don't need wrapup, so do we need to handle RONA in that case?

packages/contact-center/store/src/storeEventsWrapper.ts
```diff
@@ -235,6 +244,13 @@ class StoreWrapper implements IStoreWrapper {
     }
   };
 
+  handleTaskReject = (task: ITask, reason: string) => {
```
Kesari3008, line 247: Why do we need to pass the complete task object when only taskID is relevant in this handler?

widgets-samples/cc/samples-cc-react-app/src/App.tsx
```diff
@@ -189,9 +241,23 @@ function App() {
...
               )}
             </>
+          )}  
+
+          {showRejectedPopup && (
+            <div className="task-rejected-popup">
+              <h2>Task Rejected</h2>
+              <p>Reason: {rejectedReason}</p>
+              <select value={selectedState} onChange={(e) => setSelectedState(e.target.value)}>
+                <option value="">Select a state</option>
```
Kesari3008, line 251: We can just keep Available and Idle only as the options

Kesari3008 (changes requested):

Shreyas281299:
@adhmenon  I saw that when Im on a call and I dont do anything the UI is stuck, The incoming task and task-list were still. Can you check that scenario please.

packages/contact-center/store/src/storeEventsWrapper.ts
Shreyas281299: @Kesari3008  this is required when there is an incoming call and the agent reloads the page, now when the page reloads task:hydrate will happen. Now if we dont add this event listener and the agent doesnt pick up the call then the task:reject event will be emitted but the widgets wont listen to it and the UI will break.

adhmenon:
> @adhmenon I saw that when Im on a call and I dont do anything the UI is stuck, The incoming task and task-list were still. Can you check that scenario please.

@Shreyas281299 , checked it in multiple scenarios... I'm not seeing this actually.
Things seem to be working normally on my system... tried on firefox and chrome also.

Kesari3008 (approved):

PR #419, opened 2025-03-21
chore(tooling): add pull request checks

sreenara:
This PR adds validations to the Pull_Request CI job to ensure that the right label and title are set in the description. This is to ensure
1. Only those who have write access can trigger Actions on the PR
2. Titles match what semantic-release looks for to ensure we don't run into deployment issues

PR where this was tested: https://github.com/sreenara/widgets/pull/1
1. Created a PR against this branch to ensure the checks start
2. When the label is not added, no checks are run and everything is skipped
3. When the label is added, the title of the PR is checked and ensures that the author has set it correctly, else the remaining pipeline does not run

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -16,6 +16,55 @@ env:
   rid: ${{ github.run_id }}-${{ github.run_number }}
 
 jobs:
+  validate:
+    name: Validate Pull Request
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Validate Labels
+        run: |
+          HEAD_REPO=${{ github.event.pull_request.head.repo.full_name }}
+          BASE_REPO=${{ github.event.pull_request.base.repo.full_name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request.yml
```diff
@@ -16,8 +17,38 @@ env:
...
+    if: contains(toJson(github.event.pull_request.labels), 'validated')
+
+    steps:
+      - name: Validate PR title
+        run: |
+          TITLE="${{ github.event.pull_request.title }}"
+          echo "PR Title is: '$TITLE'"
+
+          # Pattern explanation:
+          #   ^(fix|feat|chore)   -> Must start with one of these keywords
```
mkesavan13, line 32: We should also add docs here. Docs commits can come in handy when we do sample app only changes that doesn't need a package release and this is a common case

mkesavan13 (approved):

Shreyas281299 (approved):

.github/workflows/pull-request.yml
sreenara: This is changed now.

PR #422, opened 2025-03-25
feat(widgets): added-blind-transfer

adhmenon:
## Description
* Added code for consult & transfer user interface
* Code has also been added for blind transfer working
* Unit tests for the above and some missed out cases

Vidcast - https://app.vidcast.io/share/24962b05-3a16-4b60-8c83-9333efa121f3?playerMode=vidcast

## Considerations
* Right now, this PR will get merge conflicts once the components PR is in, will resolve and update.
* I have broken up the ticket into 3 - one for transfer, consult & queues
* Test coverage is low in some areas as I was not able to get them working in time. WIP
* There is a small UI glitch in the popover, will fix it with the next PR.
* Error screens are a WIP for now until UX finalises - will be done in the next PR.

## Tests
* Tested with EXTENSION and BROWSER mode
* Tested old flows such as decline/cancel call to see that nothing was broken
* Tested it with multi login to see that state persists.


packages/contact-center/store/src/store.ts
```diff
@@ -35,6 +35,7 @@ class Store implements IStore {
   lastStateChangeTimestamp?: number;
   lastIdleCodeChangeTimestamp?: number;
   showMultipleLoginAlert: boolean = false;
+  isAgentTransferred: boolean = false;
```
adhmenon, line 38: State so we can check when the agent has been transferred.

packages/contact-center/store/src/store.types.ts
```diff
@@ -93,6 +94,7 @@ enum TASK_EVENTS {
   CONTACT_RECORDING_PAUSED = 'ContactRecordingPaused',
   CONTACT_RECORDING_RESUMED = 'ContactRecordingResumed',
   AGENT_WRAPPEDUP = 'AgentWrappedUp',
+  AGENT_BLIND_TRANSFERRED = 'AgentBlindTransferred',
```
adhmenon, line 97: As discussed, we will update above state on this event.

packages/contact-center/store/src/storeEventsWrapper.ts
```diff
@@ -10,6 +10,7 @@ import {
   IdleCode,
   IContactCenter,
   ITask,
+  BuddyDetails,
```
adhmenon, line 13: We need this for the budyd agents details.

packages/contact-center/store/src/storeEventsWrapper.ts
```diff
@@ -418,6 +437,22 @@ class StoreWrapper implements IStoreWrapper {
       });
     });
   };
+
+  getBuddyAgents = async (): Promise<Array<BuddyDetails>> => {
```
adhmenon, line 441: Method so we can extract buddy agents - queues will also be similar to this.

packages/contact-center/task/src/CallControl/CallControlCustomComponents/call-control-list-item.presentational.tsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import {ListItemBase, ListItemBaseSection, AvatarNext, Text, ButtonCircle} from '@momentum-ui/react-collaboration';
+import {Icon} from '@momentum-design/components/dist/react';
+import classnames from 'classnames';
+
+export interface CallControlListItemPresentationalProps {
```
adhmenon, line 6: This entire class is for the custom list item, right now all it consists of is an avatar, title, subtitle and a button which are all highly customisable.
Open to any suggestions to make it better.

packages/contact-center/task/src/CallControl/CallControlCustomComponents/call-control-list-item.presentational.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const {title, subtitle, buttonIcon, onButtonPress, className} = props;
+  const initials = title
+    .split(' ')
+    .map((word) => word[0])
+    .join('')
+    .slice(0, 2)
+    .toUpperCase();
+
+  return (
+    <ListItemBase className={classnames('call-control-list-item', className)} size={50} isPadded aria-label={title}>
```
adhmenon, line 24: Size is 50 by default.

packages/contact-center/task/src/CallControl/CallControlCustomComponents/call-control-popover.presentational.tsx
```diff
@@ -0,0 +1,61 @@
...
+import CallControlListItemPresentational from './call-control-list-item.presentational';
+
+export interface CallControlPopoverPresentationalProps {
+  heading: string;
+  buttonIcon: string;
+  buddyAgents: Array<{agentId: string; agentName: string; dn: string}>;
+  onAgentSelect: (agentId: string) => void;
+}
+
+const CallControlPopoverPresentational: React.FC<CallControlPopoverPresentationalProps> = ({
```
adhmenon, line 12: This is for the internal list inside the popover (ie heading, tab list and the list of agents)
We will move it once the component PR goes in as well.
It will simply render everything - it is also customisable based on trasnfer or consult.

packages/contact-center/store/src/storeEventsWrapper.ts
```diff
@@ -418,6 +437,22 @@ class StoreWrapper implements IStoreWrapper {
...
+    try {
+      const response = await this.store.cc.getBuddyAgents({
+        mediaType: 'telephony',
+        state: 'Available',
+      });
+      return response?.data?.agentList || [];
```
Kesari3008, line 447: Why do we have ? after response? In case of success, response will always be there and have data right. And even agentList is mandatory field but it can be empty array sometime.
And failure should go to catch block

packages/contact-center/task/src/helper.ts
```diff
@@ -258,6 +269,24 @@ export const useCallControl = (props: useCallControlProps) => {
...
   };
 
+  const transferCall = async (transferDestination: string, destinationType: string) => {
+    const transferPayload = {
+      to: transferDestination,
+      destinationType: destinationType,
+    };
+
+    try {
+      await currentTask.transfer(transferPayload);
```
Kesari3008, line 279: Why are we not using the response here that we receive as part of promise resolve?

packages/contact-center/task/src/task.types.ts
```diff
@@ -197,6 +197,28 @@ export interface ControlProps {
...
+   * Function to load buddy agents
+   */
+  loadBuddyAgents: () => Promise<void>;
+
+  /**
+   * Function to transfer the call to a destination.
+   * @param destination - The destination to transfer the call to.
+   * @param destinationType - The type of destination.
+   */
+  transferCall: (destination: string, destinationType: string) => void;
```
Kesari3008, line 216: We have a type in SDK for destinationType so instead of keeping it as string why don't we use that enum type

packages/contact-center/task/src/CallControl/call-control.presentational.tsx
```diff
@@ -89,24 +110,105 @@ function CallControlPresentational(props: CallControlPresentationalProps) {
...
+                      />
+                    ) : null}
+                  </PopoverNext>
+                );
+              }
+              return (
+                <TooltipNext
+                  key={index}
+                  triggerComponent={
+                    <ButtonCircle className={button.className} onPress={button.onClick} disabled={button.disabled}>
```
Kesari3008, line 197: We should add aria-label to all the Button or icons we have for accessibility purpose

Kesari3008 (changes requested):

packages/contact-center/task/src/CallControl/call-control.presentational.tsx
```diff
@@ -89,24 +110,105 @@ function CallControlPresentational(props: CallControlPresentationalProps) {
...
+                        heading={button.menuType}
+                        buttonIcon={button.icon}
+                        buddyAgents={buddyAgents}
+                        onAgentSelect={(agentId) => {
+                          setShowAgentMenu(false);
+                          setAgentMenuType(null);
+                          if (agentMenuType === 'Consult') {
+                            consultCall();
+                          } else {
+                            transferCall(agentId, 'agent');
```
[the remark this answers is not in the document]
adhmenon: Figma doesn't seem to have dial number... we only have "Agents" and "Queues" as of now.

packages/contact-center/store/src/storeEventsWrapper.ts
```diff
@@ -286,6 +299,9 @@ class StoreWrapper implements IStoreWrapper {
 
     task.on(TASK_EVENTS.AGENT_WRAPPEDUP, this.handleTaskWrapUp);
 
+    // Listen for AGENT_BLIND_TRANSFERRED event
+    task.on(TASK_EVENTS.AGENT_BLIND_TRANSFERRED, () => this.handleAgentBlindTransferred(task));
```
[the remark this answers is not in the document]
adhmenon: SDK right now does emit this event - but we will add it specifically for this. Was planning to add it with the SDK changes for queue.
We need to listen to this - this is so that we can ensure that the agent can wrapup up (since end call event is sent, right now in the widgets we have no way of knowing if it needed wrapup or not, hence we add this event to set a state then).

packages/contact-center/task/src/helper.ts
adhmenon: I think we don't need to use the response here, not sure how helpful it would be also.

packages/contact-center/store/src/storeEventsWrapper.ts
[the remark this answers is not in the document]
adhmenon: This is so we can update the state so that when we get end call event from SDK after trasnfer, we know to wrapup the task.
If I try to set the state _after_ we trasnfer (ie in helper.ts), I am worried that the asynchronous timings could mess it up. By listening to the events (which are sequential) it should work.

Also in SDK we are emitting these events as a work-around - will make it permanent with the queue PR.

packages/contact-center/task/src/task.types.ts
adhmenon: Don't we need to add code in the SDK so we can export it? Otherwise it would fail during runtime as it is an enum...

adhmenon:
Closing this PR, opened a new one https://github.com/webex/widgets/pull/423

PR #427, opened 2025-04-15
feat(cc-task): add util to get task controls visibility

rarajes2:
JIRA - [SPARK-644606](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-644606)

**Description**

This PR adds a util function to return an object with the task controls' visibility.

**Tested the following**

- Tested all the scenarios mentioned on the [Confluence](https://confluence-eng-gpk2.cisco.com/conf/x/e6EOJw) on the Agent desktop and on the widgets tested the flags values and UI visibility for calls in all the three station logins.

### DIFF
--- a/packages/contact-center/cc-components/src/components/task/CallControl/call-control.tsx
+++ b/packages/contact-center/cc-components/src/components/task/CallControl/call-control.tsx
@@ -1,11 +1,12 @@
 import React, {useEffect, useState} from 'react';
-
-import {CallControlComponentProps} from '../task.types';
-import './call-control.styles.scss';
 import {PopoverNext, SelectNext, TooltipNext, Text, ButtonCircle, ButtonPill} from '@momentum-ui/react-collaboration';
 import {Item} from '@react-stately/collections';
 import {Icon} from '@momentum-design/components/dist/react';
+
+import './call-control.styles.scss';
+import {CallControlComponentProps} from '../task.types';
 import ConsultTransferPopoverComponent from './CallControlCustom/consult-transfer-popover';
+import {getControlsVisibility} from '../../../utils/task-util';
 
 function CallControlComponent(props: CallControlComponentProps) {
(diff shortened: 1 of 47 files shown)


packages/contact-center/cc-components/src/utils/task-util.ts
```diff
@@ -0,0 +1,45 @@
+import {ITask} from '@webex/plugin-cc';
+
+/**
+ * This function determines the visibility of various controls based on the task's data.
+ * @param task The task object
+ * @returns An object containing the visibility of various controls based on the task's data
+ */
+export function getControlsVisibility(deviceType: string, featureFlags: {[key: string]: boolean}, task: ITask) {
```
mkesavan13, line 8: [taskControlUtil.txt](https://github.com/user-attachments/files/19795370/taskControlUtil.txt)

Tried to prompt and convert this to what we discussed. It took a couple of rounds and looks like we have a solution there. Do we want to try this out? This seems more descriptive.

mkesavan13 (changes requested):
Let's change the title to `feat(cc-task)` instead of `feat(*)`

packages/contact-center/cc-components/src/components/task/CallControl/call-control.tsx
```diff
@@ -30,8 +31,12 @@ function CallControlComponent(props: CallControlComponentProps) {
     loadBuddyAgents,
     transferCall,
     consultCall,
+    deviceType,
+    featureFlags,
   } = props;
 
+  const visibleCtrl = getControlsVisibility(deviceType, featureFlags, currentTask);
```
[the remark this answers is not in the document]
rarajes2: Used `useMemo` as this is not a component function

packages/contact-center/cc-components/src/utils/task-util.ts
rarajes2: As discussed, we will keep this util like this as of now. We will do the refactoring at the task level itself.

PR #433, opened 2025-04-16
feat(cc-widgets): Update Station Login UI

brain-frog:


packages/contact-center/cc-components/src/components/StationLogin/station-login.style.scss
```diff
@@ -89,3 +87,43 @@
...
+mdc-option {
+  display:block
+}
+
+.white-button {
+  border-color: var(--mdc-button-secondary-border-color) !important;
+}
+
+.cc-logout-modal {
+  width: 400px;
```
cmullenx, line 117: can you use rems consistently over px wherever px is used??

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
```diff
@@ -68,6 +70,13 @@ const StationLoginComponent: React.FunctionComponent<StationLoginComponentProps>
     }
   }, [handleContinue]);
 
+  const cancelClicked = useCallback(() => {
```
cmullenx, line 73: can you make this more verbose so its immediately clear that its handling dialog cancel button?

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
```diff
@@ -78,12 +87,21 @@ const StationLoginComponent: React.FunctionComponent<StationLoginComponentProps>
...
     (value: string) => {
-      setTeamsValue(value);
       setTeam(value);
     },
     [setTeam]
   );
 
+  const updateDialNumberLabel = (selectedOption: string): void => {
+    if (selectedOption === 'AGENT_DN') {
+      setDialNumberLabel('Dial number');
```
cmullenx, line 97: Can you make the case consistent - 1 is `Dial number`, the next line has`Dial Number` and maybe make these strings constants to avoid duplication? same for `Extension`

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
```diff
@@ -95,81 +113,106 @@ const StationLoginComponent: React.FunctionComponent<StationLoginComponentProps>
...
+            <Text tagname={'span'} type="heading-small-bold">
               Set your interaction preferences
             </Text>
           </fieldset>
           <fieldset className="fieldset">
-            <legend id="agent-login-label">Handle calls using</legend>
-            <div className="select-container">
-              <SelectNext
+            <legend id="agent-login-label">
+              Handle calls using
```
cmullenx, line 139: Why arent we using `Text` component here?

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
```diff
@@ -95,81 +113,106 @@ const StationLoginComponent: React.FunctionComponent<StationLoginComponentProps>
...
-              placeholder="Extension/Dial Number"
-              onChange={updateDN}
-              value={dialNumberValue}
-              isDisabled={isDialNumberDisabled}
-            />
-          </fieldset>
+          {isDialNumberDisabled ? (
+            <></>
+          ) : (
+            <fieldset className="fieldset">
```
cmullenx, line 173: i personally wouldn't use fieldset in this component since its such a small form, instead i would simply set the `label` property of the `Input` component. 

Just using all the momentum components for text, etc.. gets you all the correct styling for light/dark mode out of the box

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
```diff
@@ -95,81 +113,106 @@ const StationLoginComponent: React.FunctionComponent<StationLoginComponentProps>
...
             {isAgentLoggedIn ? (
-              <ButtonPill id="logoutAgent" onPress={logout} color="cancel">
+              <Button id="logoutAgent" onClick={logout} color="positive">
                 Logout
-              </ButtonPill>
+              </Button>
+            ) : (
+              <Button onClick={login}>Save & Continue</Button>
+            )}
+            {typeof contactCenterLogout === 'function' ? (
```
cmullenx, line 210: why are we doign this type check? Shouldn't truthiness check work? Typescript would show an error if someone adds this prop as anything other than a function yes? To make it more obvious its a fn, you could even call it `contactCenterLogoutFn`.

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
```diff
@@ -95,81 +113,106 @@ const StationLoginComponent: React.FunctionComponent<StationLoginComponentProps>
           </button>
         </div>
       </dialog>
+      <dialog ref={ccSignOutModalRef} className="cc-logout-modal">
```
cmullenx, line 116: Can you add a `TODO` that this dialog is tech debt and will be replaced when the momentum component is available? Otherwise it may get missed

cmullenx (commented):
minor comments

packages/contact-center/cc-components/src/components/StationLogin/station-login.style.scss
```diff
@@ -1,6 +1,5 @@
 .box.station-login {
   border-radius: 8px;
-  box-shadow: 0 2px 4px rgba(0, 0, 0, 0.1);
   padding: 20px;
```
sreenara, line 3: same question here for the px vs rem

sreenara (commented):
just a question about the px vs rem

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
```diff
@@ -95,81 +113,106 @@ const StationLoginComponent: React.FunctionComponent<StationLoginComponentProps>
...
-                {(item) => (
-                  <Item textValue={item.name} key={item.key}>
-                    <Text className="state-name" tagName={'small'}>
-                      {item.name}
-                    </Text>
-                  </Item>
-                )}
-              </SelectNext>
-              <Icon className="select-arrow-icon" name="arrow-down-bold" title="" />
+                {loginOptions.map((option, index) => {
```
mkesavan13, line 160: Two things here:
- We need to add labels for each of these types as discussed in the space
    - BROWSER = "Desktop"
    - AGENT_DN = "Dial Number"
    - EXTENSION = "Extension"
- Can we also sort this in the order [shown in Figma](https://www.figma.com/design/p5oQOpSPSaM2CzyJTc9pZV/1.-DSK-Widget---Epic-Integration--Deliverables-?node-id=3341-180275&t=9YLwsrPLnKMFEF97-0)? (Dial Number, Extension, Desktop)

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
```diff
@@ -95,81 +113,106 @@ const StationLoginComponent: React.FunctionComponent<StationLoginComponentProps>
...
-            <TextInput
-              clearAriaLabel="Clear"
-              aria-labelledby="dial-number-label"
-              placeholder="Extension/Dial Number"
-              onChange={updateDN}
-              value={dialNumberValue}
-              isDisabled={isDialNumberDisabled}
-            />
-          </fieldset>
+          {isDialNumberDisabled ? (
```
mkesavan13, line 170: Can't this be decided based on `deviceType`? Why do we need a separate boolean for this?

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
```diff
@@ -95,81 +113,106 @@ const StationLoginComponent: React.FunctionComponent<StationLoginComponentProps>
...
-                    <Text className="state-name" tagName={'small'}>
-                      {item.name}
-                    </Text>
-                  </Item>
-                )}
-              </SelectNext>
-              <Icon className="select-arrow-icon" name="arrow-down-bold" title="" />
+                {loginOptions.map((option, index) => {
+                  return (
+                    <Option key={index} value={option}>
```
mkesavan13, line 162: If it is the re-login, then the deviceType value should be selected value. At the moment with this change, it just resets to 0th value all the time I believe. 

The value should come from the store to the props. Please check the `selectedKey` in the deleted change

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
```diff
@@ -78,12 +87,21 @@ const StationLoginComponent: React.FunctionComponent<StationLoginComponentProps>
...
-      setTeamsValue(value);
       setTeam(value);
     },
     [setTeam]
   );
 
+  const updateDialNumberLabel = (selectedOption: string): void => {
+    if (selectedOption === 'AGENT_DN') {
+      setDialNumberLabel('Dial number');
+      setDialNumberPlaceholder('Dial Number');
```
mkesavan13, line 98: From the Figma, this should be **Dial number**

```suggestion
      setDialNumberPlaceholder('Dial number');
```

Also do we need two label constants for this? The text seems to be same for both. Can we move to single variable?

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
mkesavan13: Yes. It will be good to have a TASK Created as well for this in [this PF](https://jira-eng-sjc12.cisco.com/jira/browse/CAI-6238).

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
```diff
@@ -95,81 +113,106 @@ const StationLoginComponent: React.FunctionComponent<StationLoginComponentProps>
...
-          </fieldset>
+          {isDialNumberDisabled ? (
+            <></>
+          ) : (
+            <fieldset className="fieldset">
+              <legend id="dial-number-label">{dialNumberLabel}</legend>
+              <Input
+                placeholder={dialNumberPlaceholder}
+                disabled={isDialNumberDisabled}
+                value={dialNumberValue}
```
mkesavan13, line 178: The `dialNumber` value right now is local to this component and doesn't work on refresh or relogin cases. Could you please look at how `deviceType` is done (from the store) and move this as well?

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
brain-frog: In my search I had found that was a good way to make sure it was defined since it is an optional function, but truthiness check should work. I will make those updates

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
```diff
@@ -95,81 +113,106 @@ const StationLoginComponent: React.FunctionComponent<StationLoginComponentProps>
...
-                    <Text className="state-name" tagName={'small'}>
-                      {item.name}
-                    </Text>
-                  </Item>
-                )}
-              </SelectNext>
-              <Icon className="select-arrow-icon" name="arrow-down-bold" title="" />
+                {teams.map((team: {id: string; name: string}) => {
+                  return (
+                    <Option key={team.id} value={team.name}>
```
mkesavan13, line 194: I again do not see a mechanism to set the selected value which was happening earlier. You can look at the `selectedKey` in the deleted lines

mkesavan13 (changes requested):
Please look at the review comments. Also, let's change the PR title to reflect better. Something like: `fix(station-login): update UI to match figma`

It looks like the CI is failing. Please check that as well

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
brain-frog: I will add the labels as discussed
as for the sort order, I don't believe it would make sense to try and sort based on the figma as different use cases will decide if all of the options are shown in the first place. Some (like ours) won't include Desktop, some will, etc. Also as there is no sort order provided by the store and "(Dial Number, Extension, Desktop)" is not alphabetical it would end up needing to be a hard-coded sort which shouldn't be done for a dynamic array.

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
brain-frog: deviceType is undefined on initial load and does not update the conditional to show this until the login type is changed, I can try to find a workaround if you would prefer not having an additional boolean

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
brain-frog: `selectedKey` is no longer a parameter on the Select component in momentum-design, I will see how it can be replaced

packages/contact-center/cc-components/src/components/StationLogin/station-login.tsx
brain-frog: I can set it using `deviceType` but `deviceType` seems to be set on load to the first available option in the options array, is that saving it effectively?

Shreyas281299 (approved):

mkesavan13 (approved):
Had a pair programming & sit down review session with @Shreyas281299 & @brain-frog. Code looks good.

PR #538, opened 2025-10-03
feat(task): Update Outdial Dialpad UI to Match Figma

brain-frog:
# COMPLETES # https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7173

The outdial component currently does not have full functionality nor does it match the figma, and this PR addresses those areas

Adding CSS to match the figma, adding the outdial ani list for the dropdown, fixing the unit tests

### Change Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Tooling change
- [ ] Internal code refactor

- [x] The testing is done with the amplify link
< ENUMERATE TESTS PERFORMED, WHETHER MANUAL OR AUTOMATED >

The OutdialCallComponent:
- renders correctly
- updates the dial number input correctly when typing directly
- updates the dial number input correctly when clicking keypad buttons
- calls startOutdial with the correct payload upon clicking the call button
- allows special characters (* # +) from keypad
- shows the error help text when invalid characters or formats are entered
- has no ANI entry options when the list provided by the sdk is empty
- correctly sets the ani when the option is selected in the dropdown
- disables the call button when the dial number is empty

## The GAI Coding Policy And Copyright Annotation Best Practices ##
 

 
- [ ] GAI was not used (or, no additional notation is required)
- [ ] Code was generated entirely by GAI
- [ ] GAI was used to create a draft that was subsequently customized or modified
- [ ] Coder created a draft manually that was non-substantively modified by GAI (e.g., refactoring was performed by GAI on manually written code)
- [x] Tool used for AI assistance (GitHub Copilot / Other - specify)
  - [x] Github Copilot
  - [ ] Other - Please Specify
- [x] This PR is related to
  - [x] Feature
  - [ ] Defect fix
  - [ ] Tech Debt
  - [ ] Automation

### Checklist before merging

- [x] I have updated the testing document
- [x] I have tested the functionality with amplify link

---

Make sure to have followed the [contributing guidelines](https://github.com/webex/webex-js-sdk/blob/master/CONTRIBUTING.md#submitting-a-pull-request) before submitting.


packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -2,50 +2,160 @@ import React, {useState} from 'react';
...
+ * @property startOutdial - Function to initiate the outdial call with the entered destination number.
+ * @property dialNumberRegex - Optional regular expression provided by the CC Store for validating destination number format.
+ */
 const OutdialCallComponent: React.FunctionComponent<OutdialCallComponentProps> = (props) => {
-  const {startOutdial} = props;
+  const {startOutdial, dialNumberRegex} = props;
+
+  // State Hooks
   const [destination, setDestination] = useState('');
+  const [error, setError] = useState('');
```
cmullenx, line 48: can you name this something like `isValidNumber` to make it more clear, you will likely need to implement an error state for if something in the component goes wrong so these consts should be reserved for that

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -2,50 +2,160 @@ import React, {useState} from 'react';
...
-            </div>
-            <button className="out-dial-call-btn" onClick={() => startOutdial(destination)}>
-              <svg xmlns="http://www.w3.org/2000/svg" height="24" viewBox="0 0 24 24" width="24">
-                <path d="M6.62 10.79a15.053 15.053 0 006.59 6.59l2.2-2.2a1 1 0 011.11-.27c1.12.45 2.33.69 3.58.69a1 1 0 011 1v3.5a1 1 0 01-1 1C10.29 21 3 13.71 3 4.5a1 1 0 011-1h3.5a1 1 0 011 1c0 1.25.24 2.46.69 3.58a1 1 0 01-.27 1.11l-2.2 2.2z" />
-              </svg>
-            </button>
-          </div>
-        </fieldset>
-      </section>
+    <div className="keypad">
```
cmullenx, line 100: can you make the html more semantic? like you can use section, input type of `tel`, using `ul` and `li` for the buttons, etc..?

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -2,50 +2,160 @@ import React, {useState} from 'react';
...
-  // Function to press a key on the outdial keypad.
-  const handelKeyPress = (value: string) => {
-    setDestination((prev) => prev + value);
+  /**
+   * handleKeyPress
+   * @param value The key value pressed
+   * Appends the pressed key to the destination input field
+   */
+  const handleKeyPress = (value: string) => {
+    updateOutboundNumber({target: {value: destination + value}} as React.ChangeEvent<HTMLInputElement>);
```
cmullenx, line 96: why not just pass in a string? this is harder to reason about

cmullenx (commented):
what happens if you fail to fetch the outdial ani list? what happens if the call fails? Tests?

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -2,50 +2,160 @@ import React, {useState} from 'react';
...
     // Allow only valid input that is digits, #, *, and +
     const VALID_KEYPAD_CHARS = /[\d#*+]/g;
-    const filteredValue = e.target.value.match(VALID_KEYPAD_CHARS)?.join('') || '';
-    setDestination(filteredValue);
+    const inputValue = e.target.value;
+    const filteredValue = inputValue.match(VALID_KEYPAD_CHARS)?.join('') || '';
+    setDestination(inputValue);
+
+    // Validate the input format
+    const regexForDn = new RegExp(dialNumberRegex ?? '1[0-9]{3}[2-9][0-9]{6}([,]{1,10}[0-9]+){0,1}');
```
cmullenx, line 81: also create this regex outside the this fn and useMemo, else function is created on ever change event

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
brain-frog: Great point! It was originally only being used by the onChange so the event made sense to use, but string will be more readable

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -1,52 +1,166 @@
...
+
+  // useEffect and useState to allow for async fetching of outdial ANI entries
+  useEffect(() => {
+    // Give Select an empty list if outdial ANI entries are not provided
+    const updateList = async () => {
+      try {
+        const result = await getOutdialANIEntries();
+        setOutdialANIList(result);
+      } catch (error) {
+        console.error('Error fetching outdial ANI entries:', error);
```
cmullenx: should this be using logger and not console.log?

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -1,52 +1,166 @@
...
 
-  // Function to press a key on the outdial keypad.
-  const handelKeyPress = (value: string) => {
-    setDestination((prev) => prev + value);
+  /**
+   * handleKeyPress
+   * @param value The key value pressed
+   * Appends the pressed key to the destination input field
+   */
+  const handleKeyPress = (value: string) => {
```
cmullenx: can you change to handleOnClick since its called onClick?

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -1,52 +1,166 @@
...
+        <Tab
+          active={true}
+          text={OutdialStrings.DIALPAD_LABEL}
+          iconName={'dialpad-bold'}
+          tabId="dialpad-tab"
+          variant="pill"
+          aria-controls="dialpad-panel"
+        ></Tab>
+      </header>
+      <Input
```
cmullenx, line 76: does the input component not take a type? if it does, seems like this should be type number

packages/contact-center/task/src/helper.ts
```diff
@@ -868,7 +867,27 @@ export const useOutdialCall = (props: useOutdialCallProps) => {
...
   };
 
+  /**
+   * Fetches the Outdial ANI entries for the current agent.
+   * @returns An array of Outdial ANI entries.
+   */
+  const getOutdialANIEntries = async () => {
+    try {
+      const agentProfile = cc.agentConfig;
+      const outdialANIId = agentProfile?.outdialANIId || '';
```
cmullenx: if there's not an outdial ani, seems we should log it and then not make the call to getOutdialAniEntries

packages/contact-center/task/src/helper.ts
```diff
@@ -842,15 +841,15 @@ export const useCallControl = (props: useCallControlProps) => {
 export const useOutdialCall = (props: useOutdialCallProps) => {
   const {cc, logger} = props;
 
-  const startOutdial = (destination: string) => {
+  const startOutdial = (destination: string, origin: string = undefined) => {
```
cmullenx, line 900: when there is not an origin, seems like we shouldn't make the call to outdial and instead log an error

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.style.scss
```diff
@@ -1,67 +1,43 @@
...
-    text-align: center;
-    font-size: 1.125rem;
-    margin-bottom: 0.625rem;
-    background: var(--mds-color-theme-background-primary-normal);
-    border: 1px solid var(--mds-color-theme-outline-secondary-normal);
-    border-radius: 0.25rem;
-    color: var(--mds-color-theme-text-primary-normal);
   }
 
+  .button {
```
cmullenx: can you give these less generic names because if there are other elements using class '.button' this will override them correct?

Also does button size prop not work? couldnt you set the size of the button with props?

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.style.scss
```diff
@@ -1,67 +1,43 @@
...
+  mdc-input {
+    padding-bottom: 0; // default is 1 rem, 1.5 rem needed but provided by .keys
+  }
+  
+  #outdial-tablist {
+    margin-bottom: 1rem;
+  }
 
-  input {
+  .input {
```
cmullenx: same thing here as button comment

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
brain-frog: yes, let me add logger as a property of outdial call

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
brain-frog: none that I saw in the documentation or in the node modules declaration

packages/contact-center/task/src/helper.ts
brain-frog: thank you and good catch, that was leftover from before the sdk updates were made

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.style.scss
brain-frog: Replacing this with the size prop, thank you

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.style.scss
brain-frog: replacing the name here to be more specific, thank you

packages/contact-center/task/src/helper.ts
brain-frog: after checking your SDK PR as well as agent desktop it does not seem that an origin is required for an outdial call
if that's incorrect I can add a logger error and validation for it

mkesavan13 (changes requested):
- It looks like the sync from `ccwidgets` branch did not happen properly and I see those changes appearing in this PR also
- Let us bring back the existing PR description template and fill it in. It is a leadership ask to have the GAI policy as part of all the PRs and filled in appropriately.
- Kindly record a vidcast of the changes to showcase and attach it to the description
- Add a brief note of all the scenarios that were tested

brain-frog:
vidcast showing functionality: https://app.vidcast.io/share/d0be837f-11f2-4289-ac7c-73d262a78c33

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -1,52 +1,166 @@
...
+  const [outdialANIList, setOutdialANIList] = useState<OutdialANIEntry[]>([]);
+
+  // Validate the input format using regex from agent desktop
+  const regExForDnSpecialChars = useMemo(
+    () => new RegExp('^[+1][0-9]{3,18}$|^[*#][+1][0-9*#:]{3,18}$|^[0-9*#]{3,18}$'),
+    []
+  );
+
+  // useEffect and useState to allow for async fetching of outdial ANI entries
+  useEffect(() => {
```
Kesari3008, line 34: What about pulling entries for address book ? When the outdial component loads, we have 3 options in Agent desktop currently: Dialpad, searching from address book and organization search so since address book is available now would be good to add that as well here

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -1,52 +1,166 @@
...
+ */
+interface OutdialANIEntry {
+  organizationId?: string;
+  id?: string;
+  version?: number;
+  name: string;
+  number: string;
+  createdTime?: number;
+  lastUpdatedTime?: number;
+}
```
mkesavan13: Can we move this into a types file like the other components?

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -1,52 +1,166 @@
...
+      try {
+        const result = await getOutdialANIEntries();
+        setOutdialANIList(result);
+      } catch (error) {
+        logger?.error('Error fetching outdial ANI entries:', error);
+        setOutdialANIList([]);
+      }
+    };
+    updateList();
+  }, [getOutdialANIEntries]);
```
mkesavan13: Are we expecting the `getOutdialANIEntries` method to change often? Why is that added as a dependency here?

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -1,52 +1,166 @@
...
-    // Allow only valid input that is digits, #, *, and +
-    const VALID_KEYPAD_CHARS = /[\d#*+]/g;
-    const filteredValue = e.target.value.match(VALID_KEYPAD_CHARS)?.join('') || '';
-    setDestination(filteredValue);
+  /**
+   * validateOutboundNumber
+   * @param e The input change event
+   * If the input is invalid, sets an error message on dialnumber input
+   */
+  const validateOutboundNumber = (value: string) => {
```
mkesavan13: The param name in the doc and the param name in the actual method below doesn't match

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -1,52 +1,166 @@
...
+
+  // useEffect and useState to allow for async fetching of outdial ANI entries
+  useEffect(() => {
+    // Give Select an empty list if outdial ANI entries are not provided
+    const updateList = async () => {
+      try {
+        const result = await getOutdialANIEntries();
+        setOutdialANIList(result);
+      } catch (error) {
+        logger?.error('Error fetching outdial ANI entries:', error);
```
Kesari3008: Why is logger optional ? This is coming from store and it seems from other components where this is being used, this is not an optional parameter so we can avoid ? here.

packages/contact-center/cc-components/src/components/task/task.types.ts
```diff
@@ -477,14 +477,34 @@ export type CallControlComponentProps = Pick<
   | 'consultTransferOptions'
 >;
 
+export type OutdialAniEntry = {
```
mkesavan13, line 480: Why do we have this in two places? I saw one in the component itself. Ideally it should be in task types and get imported in the out dial UI component (cc-components)

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
Kesari3008: Add log context: module and method name as we are doing for other logger statements

packages/contact-center/store/src/store.types.ts
```diff
@@ -18,6 +18,10 @@ import {
   ContactServiceQueueSearchParams,
   AddressBook,
 } from '@webex/contact-center';
+import {
+  OutdialAniEntriesResponse,
+  OutdialAniParams,
+} from 'node_modules/@webex/contact-center/dist/types/services/config/types';
```
mkesavan13, line 24: The right way to do this would be to export the type in SDK and then bring it. Pulling it in from node_modules would affect if tomorrow we change the directory structure in the SDK

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -1,52 +1,166 @@
...
+      <header role="tablist" id="outdial-tablist">
+        <Tab
+          active={true}
+          text={OutdialStrings.DIALPAD_LABEL}
+          iconName={'dialpad-bold'}
+          tabId="dialpad-tab"
+          variant="pill"
+          aria-controls="dialpad-panel"
+        ></Tab>
+      </header>
```
mkesavan13: Do we need this header here or we should keep that as part of Epic App? Like how station login is a plain component in widgets but is put inside a Modal in epic app?

mkesavan13 (changes requested):
1. I see neither the enumerated tests nor the vidcast showing if a task actually came in
2. The out dial in the Agent Desktop always is auto answered. Did we check for whether it was auto accepted and put in the task list in the widgets?
3. Did we try to test this scenario in multi-login scenarios? Epic is heavily dependent on multi-login and it is essential to test that
4. w.r.t the components UT for out dial, we do not have the util methods like `regExForDnSpecialChars`, `validateOutboundNumber`, `handleOnClick` tested. Which is why in other components we have split them into a separate util file to test it separately. We should do the same here

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
brain-frog: the header and tabs are to allow for the switch between addressbook and keypad. If you plan to have addressbook as a separate component, i can remove this and have it as a part of Epic App

packages/contact-center/store/src/store.types.ts
brain-frog: @mkesavan13 I apologize i was following the example of the other imports in the file. These are indeed being exported by the sdk [as seen in Christina's PR](https://github.com/webex/webex-js-sdk/pull/4513/files#diff-abf4bb5b522f162ed6dbabfac2c5ddd2d8767207a298845543fba0ff867ba9ba), yet when I was trying to import them I was seeing no exported type was found

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
brain-frog: I'll remove it

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
brain-frog: replaced with the import from task types

brain-frog:
@mkesavan13 
For your first three comments, I will do some additional manual testing

> w.r.t the components UT for out dial, we do not have the util methods like regExForDnSpecialChars, validateOutboundNumber, handleOnClick tested. Which is why in other components we have split them into a separate util file to test it separately. We should do the same here

These functions are being unit tested by the tests that change the inputs and simulate clicks. They can be moved to a utils file if that is your preference

packages/contact-center/store/src/store.types.ts
mkesavan13: Oh, I see.

@Shreyas281299 - Could you please check why aren't we able to import directly? Also why do we have other `node_modules` based imports there?

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
mkesavan13: Oh. Good catch there. As we do not have the address book at the moment, does it make sense to even have the tab? Should we consider or decide whether to have a single tab based component or multiple components when we have the address book coming up?

Yes, it won't align with the Figma at the moment, but we will have a better look without the tab button. Right now it looks odd. Kindly check with Soundarya about this.

mkesavan13:
> These functions are being unit tested by the tests that change the inputs and simulate clicks. They can be moved to a utils file if that is your preference

Did we try and fail the tests to see if the test to read the screen for errors is right? If so, we are good there. @Shreyas281299  - I would like for you to verify this also once

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
brain-frog: I have asked Soundarya and for now we can remove the tab and header. Whether we bring it back here or in Epic App in the future I will leave to your team to decide if Addressbook and OutdialCall will be a single or separate components

packages/contact-center/cc-components/src/components/task/task.types.ts
```diff
@@ -477,14 +477,34 @@ export type CallControlComponentProps = Pick<
...
   /**
    * Function to start outdial call.
    */
-  startOutdial: (destination: string) => void;
+  startOutdial: (destination: string, origin: string | undefined) => void;
+
+  /**
+   * Function to get a list of Outdial ANI entries.
+   */
+  getOutdialANIEntries?: () => Array<OutdialAniEntry>;
```
Kesari3008: Why is this method an optional prop ?

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
```diff
@@ -1,52 +1,166 @@
...
+ *
+ * @property {string} organizationId - The organization ID the ANI is associated with.
+ * @property {string} id - ID of this contact center ANI entry.
+ * @property {number} version - The version number of the ANI entry.
+ * @property {string} name - The name assigned to the ANI entry.
+ * @property {string} number - The phone number associated with the ANI entry.
+ * @property {number} createdTime - The timestamp(in epoch milliseconds) when the ANI entry was created.
+ * @property {number} lastUpdatedTime - The timestamp(in epoch milliseconds) when the ANI entry was last updated.
+ */
+interface OutdialANIEntry {
```
Kesari3008: We already have a similar type created in types file then we do we need this interface to be defined here ? Why can;t we import the type created from types file and use that

brain-frog:
@mkesavan13 after further reviewing the functionality you suggested could be moved to a utils I don't believe it would make sense to move them. 
1. `regExForDnSpecialChars` uses the `useMemo` hook which should be kept within react components
2. `validateOutboundNumber` and `handleOnClick` both directly use `useState` hook setters which would not make sense to add the extra step of making them parameters to move to a utils file

They are both still tested by existing unit test simulations. Please let me know if you have additional thoughts.

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
Kesari3008: Having dependency on this method will not have any impact and based on the implementation it doesn;t look like we will have to invoke getOutdialANIEntries() again and again. When OutdialCall component loads, during the component mount , entries will be fetched and we need to run it again so we can keep this dependency array empty

packages/contact-center/cc-components/src/components/task/task.types.ts
```diff
@@ -477,14 +477,34 @@ export type CallControlComponentProps = Pick<
...
+
 /**
  * Interface representing the properties for OutdialCall component.
  */
 export interface OutdialCallProps {
   /**
    * Function to start outdial call.
    */
-  startOutdial: (destination: string) => void;
+  startOutdial: (destination: string, origin: string | undefined) => void;
```
Kesari3008: We can just make origin option by adding ? instead of specifically saying undefined

packages/contact-center/store/src/store.types.ts
```diff
@@ -51,8 +55,10 @@ interface IContactCenter {
   agentConfig?: {
     regexUS: RegExp | string;
     agentId: string;
+    outdialANIId: string;
```
Kesari3008, line 58: We will need another value to be read from agentConfig which Outdial flag: isOutboundEnabledForAgent. If outdial is enabled from control hub, this value is received and outdial component should be disabled if this flag is set to false

widgets-samples/cc/samples-cc-react-app/src/App.tsx
```diff
@@ -827,7 +827,16 @@ function App() {
                         </section>
                       </div>
                     )}
-                    {selectedWidgets.outdialCall && <OutdialCall />}
+                    {selectedWidgets.outdialCall && (
+                      <div className="box">
+                        <section className="section-box">
+                          <fieldset className="fieldset">
+                            <legend className="legend-box">Outdial Call</legend>
+                            <OutdialCall />
```
Kesari3008, line 835: Hide this component behind the feature flag mentioned for agentConfig

packages/contact-center/task/src/helper.ts
```diff
@@ -922,7 +922,30 @@ export const useOutdialCall = (props: useOutdialCallProps) => {
...
+  const getOutdialANIEntries = async () => {
+    try {
+      const agentProfile = cc.agentConfig;
+      const outdialANIId = agentProfile?.outdialANIId;
+      if (!outdialANIId) {
+        throw Error('No OutdialANI Id received.');
+      }
+      const result = await cc.getOutdialAniEntries({outdialANI: outdialANIId});
+      return result;
+    } catch (error) {
```
Kesari3008, line 939: Why is it that in some places error type is mentioned but in some places it isnt ? Would be good to check if error type would be valid to mention here

packages/contact-center/task/tests/helper.ts
```diff
@@ -2690,6 +2690,7 @@ describe('useOutdialCall', () => {
   const ccMock = {
     ...mockCC,
```
Kesari3008: Test for getOutdialAniEntries methodf should be added here. Success and failed case

Kesari3008 (changes requested):

Kesari3008:
Please add a vidcast for the testing done in PR description.
Also won't the presentational component changes need to be updated in snapshot tests we have ?
Pipeline checks are not running, please enable

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
brain-frog: Address book is not part of the scope of my current story, that can be added at a later date. This PR is to update the Outdial Dialpad

packages/contact-center/cc-components/src/components/task/OutdialCall/outdial-call.tsx
brain-frog: @Kesari3008 The optional chaining on logger is required as it will cause unit tests to fail and cause warnings of logger being undefined when running the application otherwise. Logger is not always immediately initialized on first render and throws `TypeError: Cannot read properties of undefined` in unit tests even when mocked.

packages/contact-center/cc-components/src/components/task/task.types.ts
brain-frog: I did not believe it was always required as an origin is not always required for `startOutdial`
I can remove the ?

packages/contact-center/task/src/helper.ts
brain-frog: I referenced other `try/catch` in the project and didn't deem it necessary to mention exact type. error is typecast 8 times out of 157 catch statements.
I will check the SDK and see if a specific error is expected

packages/contact-center/cc-components/tests/components/task/OutdialCall/out-dial-call.tsx
```diff
@@ -1,92 +1,144 @@
...
+    {name: 'name 2', number: '2'},
+  ]);
+  const mockLogger: ILogger = {
+    log: jest.fn(),
+    info: jest.fn(),
+    warn: jest.fn(),
+    error: jest.fn(),
+    trace: jest.fn(),
+  };
+  const KEY_LIST = ['1', '2', '3', '4', '5', '6', '7', '8', '9', '*', '0', '#'];
```
Shreyas281299: Can we use the existing constants

packages/contact-center/cc-components/tests/components/task/OutdialCall/out-dial-call.tsx
```diff
@@ -1,92 +1,144 @@
...
+    expect(screen.getByTestId('outdial-call-button')).toBeInTheDocument();
   });
 
-  it('updates input value when typing directly', () => {
+  it('updates input value when typing directly', async () => {
     render(<OutdialCallComponent {...props} />);
-    const input = screen.getByPlaceholderText('Enter number to dial');
-    fireEvent.change(input, {target: {value: '123'}});
-    expect(input).toHaveValue('123');
+    const input = await screen.findByTestId('outdial-number-input');
```
Shreyas281299, line 95: This way of rendering components isnt correct. Please check any other component for how to render the component properly we need to await, when we wait for the component to be rendered it also adds the shadow-root otherwise we dont get that. Also please add prop checks (again you'll be able to find examples in other components)

packages/contact-center/store/src/store.types.ts
Shreyas281299: Il check this.

packages/contact-center/task/src/helper.ts
```diff
@@ -922,7 +922,30 @@ export const useOutdialCall = (props: useOutdialCallProps) => {
...
+  /**
+   * Fetches the Outdial ANI entries for the current agent.
+   * @returns An array of Outdial ANI entries.
+   */
+  const getOutdialANIEntries = async () => {
+    try {
+      const agentProfile = cc.agentConfig;
+      const outdialANIId = agentProfile?.outdialANIId;
+      if (!outdialANIId) {
+        throw Error('No OutdialANI Id received.');
```
Shreyas281299, line 935: We should not be doing this. If this error is ever triggered it will break the component, I think we should just be logging this as this would be a widget's issue. What do you think?

packages/contact-center/task/tests/helper.ts
```diff
@@ -2786,6 +2787,309 @@ describe('useOutdialCall', () => {
     expect(ccMock.startOutdial).not.toHaveBeenCalled();
     expect(logger.info).not.toHaveBeenCalled();
   });
+
+  describe('getOutdialAniEntries', () => {
+    const mockAniEntries = [
```
Shreyas281299: Please move these to test-fixtures package, and then we can use the mocks from there. Feel free to structure the folders the way you want in the test-fixtures for outdial

Shreyas281299 (changes requested):

packages/contact-center/store/src/store.types.ts
cmullenx: the ability to show/hide this widget will be in our epic project and not in widgets

packages/contact-center/cc-components/tests/components/task/OutdialCall/out-dial-call.tsx
brain-frog: the difference I am seeing here is whether to get screen by de-structuring the render response or from the import, of which [using it by import is recommended](https://stackoverflow.com/questions/61482418/screen-vs-render-queries)

packages/contact-center/task/src/helper.ts
brain-frog: Sure, I can change it to a logging statement so it does not throw a breaking error

packages/contact-center/task/tests/helper.ts
brain-frog: I will move the frequently used constants and keep the test case specific ones inside the applicable tests

packages/contact-center/task/src/helper.ts
brain-frog: Looking at the sdk changes, I believe calling `getOutdialAniEntries` without a defined outdialANI will throw an error, which would still break the component right? Is it worth logging the error and letting that happen instead of throwing the error prior to the call?

packages/contact-center/cc-components/tests/components/task/OutdialCall/out-dial-call.tsx
brain-frog: The reason I redefined it is due to frequent errors when trying to import KEY_LIST, it is unable to find the module

packages/contact-center/cc-components/tests/components/task/OutdialCall/out-dial-call.tsx
Shreyas281299: https://github.com/webex/widgets/blob/572450a24dd970226e665df750142958628c0715/packages/contact-center/cc-components/tests/components/task/IncomingTask/incoming-task.tsx#L69

We wanna do something like this. You can try rendering in both the ways and print the component to see the difference

packages/contact-center/cc-components/tests/components/task/OutdialCall/out-dial-call.tsx
Shreyas281299: Before importing we have to build the parent package.

brain-frog:
Vidcast showcasing multi-login and desktop call: https://app.vidcast.io/share/be4ad199-98ce-40a7-9007-c70bc4a85e93

packages/contact-center/cc-components/tests/components/task/OutdialCall/out-dial-call.snapshot.tsx
```diff
@@ -0,0 +1,149 @@
...
+    container.querySelectorAll('[id^="mdc-input"]').forEach((el) => el.removeAttribute('id'));
+    expect(container).toMatchSnapshot();
+  });
+
+  it('does not allow empty input', async () => {
+    const {container} = render(<OutdialCallComponent {...props} />);
+    await screen.findByTestId('outdial-call-button');
+    // Remove IDs to avoid snapshot issues with dynamic IDs
+    container.querySelectorAll('[id^="mdc-input"]').forEach((el) => el.removeAttribute('id'));
+    expect(container).toMatchSnapshot();
```
Shreyas281299, line 147: This is awesome. Thanks @brain-frog

packages/contact-center/cc-components/tests/components/task/OutdialCall/out-dial-call.tsx
```diff
@@ -1,92 +1,187 @@
...
+    logger: mockCC.LoggerProxy,
+    startOutdial: jest.fn(),
+    getOutdialANIEntries: jest.fn().mockResolvedValue([
+      {name: 'name 1', number: '1'},
+      {name: 'name 2', number: '2'},
+    ]),
+  };
+  describe('renders the component correctly, should render:', () => {
+    it('article container', async () => {
+      render(<OutdialCallComponent {...props} />);
```
Shreyas281299, line 36: I still see that we are not awaiting for the rendering to happen. Did you check if the props when we put an await is the same when we dont?

packages/contact-center/cc-components/tests/components/task/OutdialCall/out-dial-call.snapshot.tsx
```diff
@@ -0,0 +1,149 @@
...
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+    jest.restoreAllMocks();
+  });
+
+  describe('Rendering', () => {
+    it('renders the component correctly', async () => {
+      const {container} = render(<OutdialCallComponent {...props} />);
```
Shreyas281299, line 35: Here also we should be awiating for the render.

packages/contact-center/cc-components/tests/components/task/OutdialCall/out-dial-call.tsx
brain-frog: Yes I have verified this still works as intended, and it was something I showed you when we called

Shreyas281299 (approved):

mkesavan13 (approved):

PR #644, opened 2026-03-09
docs(ai-docs): task refactor migration overview (PR 1/4)

akulakum:
# COMPLETES #https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7714

## This pull request addresses

The **CC SDK task-refactor branch** has completed a major architectural shift — replacing ad-hoc task state management with a **state-machine-driven architecture**. The SDK now computes UI control states (`task.uiControls`) internally and exposes them via a getter and a `task:ui-controls-updated` event, eliminating the need for widgets to derive control visibility from raw task data.

This is **PR 1 of 4** in the task refactor migration documentation series. It provides the **single entry-point overview document** — the architectural comparison, execution order, SDK prerequisites, key types, files affected, and SDK reference.

## by making the following changes

Added 1 consolidated migration overview document under `packages/contact-center/ai-docs/migration/`:

- **migration-overview.md**: Single entry point with old vs new architecture diagram, `TaskUIControls` structure, execution order (6 migration docs to follow in sequence), SDK pending exports, key types from SDK, CC Widgets files affected, and CC SDK reference

Also updated:
- **AGENTS.md** (root): Updated migration docs reference to point to `migration-overview.md`

## CC SDK Task-Refactor Branch Reference

> **These migration docs are driven by the CC SDK `task-refactor` branch.** Reviewers should evaluate comments against the **new SDK behavior**, not the old widget logic.

**Repo:** [webex/webex-js-sdk (task-refactor)](https://github.com/webex/webex-js-sdk/tree/task-refactor)

**Key SDK files:**
| File | Purpose |
|------|----------|
| `uiControlsComputer.ts` | Computes `TaskUIControls` from `TaskState` + `TaskContext` |
| `Task.ts` | Exposes `task.uiControls` getter and `task:ui-controls-updated` event |
| `constants.ts` | `TaskState` and `TaskEvent` enums |

### Change Type
- [x] Documentation update

### Related PRs
- **PR 2/4**: Store Layer — [#646](https://github.com/webex/widgets/pull/646)
- **PR 3/4**: Widget Layer — [#647](https://github.com/webex/widgets/pull/647)
- **PR 4/4**: Task List & Component Layer — [#648](https://github.com/webex/widgets/pull/648)

## The following scenarios were tested
- Documentation-only change — no functional testing required

## The GAI Coding Policy And Copyright Annotation Best Practices
- [x] GAI was used to create a draft that was subsequently customized or modified
- Tool used for AI assistance: Other - Cursor
[...]

PR #646, opened 2026-03-11
docs(ai-docs): task refactor migration — store layer (PR 2/4)

akulakum:
# COMPLETES #https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7714

This is **PR 2 of 4** in the task refactor migration documentation series. It covers the **store layer** changes — event wiring simplification and store utility function migration.

Added 2 store layer migration documents under `packages/contact-center/ai-docs/migration/`:

- **store-event-wiring-migration.md**: Documents how 30+ event handler subscriptions in `storeEventsWrapper.ts` change:
   -  Switch event names to SDK TASK_EVENTS enum (delete local copy, import from @webex/contact-center)
   - Keep refreshTaskList() in all existing handlers (store doesn't observe task.data directly; removal is a future widget optimization)
   - New TASK_UI_CONTROLS_UPDATED subscription to trigger widget re-renders
   - Replace isDeclineButtonEnabled store property with task.uiControls.decline.isEnabled
   - Fix TASK_CONSULT_END wiring to existing (dead) handleConsultEnd method
   - 3 renamed events (task-refactor: AGENT_* → TASK_) + 2 pre-existing name mismatches (recording events)
   - 4 store-only enum members to delete (no SDK equivalent)
   - Complete SDK event reference (Task.ts + Voice.ts) with string values
   - Definitive event registration table split by: no change (22), changes required (7), new additions (1)
   - Pre-existing bugs documented: dead handleConsultEnd, handleTaskRemove listener mismatch, test gap (CAI-7758)
   - Store consult state observables retained for now (widget timer dependency)
   - Before/after code examples for registerTaskEventListeners and handleAutoAnswer

- **store-task-utils-migration.md**: Identifies ~15 utility functions and their migration fate:
   - REMOVE (5 — SDK handles via uiControls): getConsultStatus, getIsConferenceInProgress, getConferenceParticipantsCount, getIsCustomerInCall, getIsConsultInProgress
   - KEEP (7 — widget-layer concerns): isIncomingTask, getTaskStatus, getConferenceParticipants, isInteractionOnHold, findMediaResourceId, findHoldTimestamp, findHoldStatus
   - REVIEW (4 — may simplify): isSecondaryAgent, isSecondaryEpDnAgent, getConsultMPCState, setmTypeForEPDN
   - Constants to delete (phased — only after dependent functions rewritten) / keep tables with ordering constraints
   - TASK_STATE_CONSULT dual-use documented (task state vs media-type sentinel in findHoldStatus)
   - Gotcha: TaskState.CONSULT_INITIATING vs CONSULTING distinction
   - getControlsVisibility + 22 visibility functions deletion scope with critical notes on derived-state consumers and feature-flag gating overlay
   - Before/after: getTaskStatus rewritten to use task.uiControls (preserving return contract for barrel exports)
   - Pre-migration step: confirm downstream usage (Epic) before removing exports
   - findHoldTimestamp dual-signature note (store ITask vs task-util Interaction)

## CC SDK Task-Refactor Branch Reference

> **These migration docs are driven by the CC SDK `task-refactor` branch.** Reviewers should evaluate comments against the **new SDK behavior**, not the old widget logic.

### Change Type
- [x] Documentation update

### Related PRs
- **PR 1/4**: Foundation & Overview — [#644](https://github.com/webex/widgets/pull/644)
- **PR 3/4**: Widget Hooks — [#647](https://github.com/webex/widgets/pull/647)
- **PR 4/4**: Task List & Component Layer — [#648](https://github.com/webex/widgets/pull/648)

- Documentation-only change — no functional testing required

## The GAI Coding Policy And Copyright Annotation Best Practices
- [x] GAI was used to create a draft that was subsequently customized or modified
- Tool used for AI assistance: Other - Cursor
- This PR is related to: Tech Debt

### Checklist before merging


akulakum:
@codex review

akulakum:
@codex review

akulakum:
@codex review

akulakum:
@codex review

akulakum:
@codex review

akulakum:
@codex review

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,486 @@
...
+| `AGENT_OFFER_CONTACT` | `'AgentOfferContact'` | `TASK_OFFER_CONTACT` | `'task:offerContact'` |
+| `CONTACT_RECORDING_PAUSED` | `'ContactRecordingPaused'` | `TASK_RECORDING_PAUSED` | `'task:recordingPaused'` |
+| `CONTACT_RECORDING_RESUMED` | `'ContactRecordingResumed'` | `TASK_RECORDING_RESUMED` | `'task:recordingResumed'` |
+
+### 4 Store-Only Enum Members (no SDK equivalent — delete)
+
+| Widget Enum Member | Value | Note |
+|---|---|---|
+| `TASK_UNHOLD` | `'task:unhold'` | SDK uses `TASK_RESUME` instead |
+| `TASK_CONSULT` | `'task:consult'` | No SDK equivalent; consult flow uses multiple events |
```
mkesavan13, line 126: The spec file in this path: [packages/contact-center/task/ai-docs/widgets/CallControl/ARCHITECTURE.md](https://github.com/webex/widgets/tree/next/packages/contact-center/task/ai-docs/widgets/CallControl/ARCHITECTURE.md) mentions about this event. That needs to be handled as well.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,486 @@
...
+| `TASK_RECORDING_RESUME_FAILED` | `'task:recordingResumeFailed'` | Evaluate for error handling |
+| `TASK_EXIT_CONFERENCE` | `'task:exitConference'` | Evaluate for conference flow |
+| `TASK_TRANSFER_CONFERENCE` | `'task:transferConference'` | Evaluate for conference flow |
+| `TASK_CLEANUP` | `'task:cleanup'` | SDK internal — likely no widget action |
+
+**Action:** Delete the local `TASK_EVENTS` enum from `store/src/store.types.ts` and import from SDK instead. SDK's `TASK_EVENTS` already includes all needed events including `TASK_UI_CONTROLS_UPDATED`.
+
+### Pre-existing Bug: Event Name Mismatches
+
+The 5 renamed events above are currently hardcoded in `store.types.ts` with a TODO comment: `// TODO: remove this once cc sdk exports this enum`. During migration, replace the entire local enum with SDK's exported `TASK_EVENTS` enum.
```
mkesavan13: Do we know if the SDK is exporting this already?

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,486 @@
...
+| 23 | `TASK_CONFERENCE_STARTED` | `handleConferenceStarted` | (same as #20) |
+| 24 | `TASK_CONFERENCE_TRANSFERRED` | `refreshTaskList` | Re-fetch all tasks |
+| 25 | `TASK_CONFERENCE_TRANSFER_FAILED` | `refreshTaskList` | Re-fetch all tasks |
+| 26 | `TASK_POST_CALL_ACTIVITY` | `refreshTaskList` | Re-fetch all tasks |
+| 27 | `TASK_MEDIA` | `handleTaskMedia` | Browser-only: `setCallControlAudio(new MediaStream([track]))` |
+
+### Pre-existing Bugs in Old Code
+
+**Bug 1: `handleConsultEnd` is dead code.**
+A `handleConsultEnd` method exists (resets `isQueueConsultInProgress`, `currentConsultQueueId`, `consultStartTimeStamp`) but `TASK_CONSULT_END` is wired to `refreshTaskList()` instead. The method's consult state cleanup never runs.
```
mkesavan13: While this is a dead code, it is being used in the UT. Which means the UT is dead too. The spec needs a plan to fix that too

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,486 @@
...
+
+**Bug 2: `handleTaskRemove` listener mismatch.**
+`registerTaskEventListeners` wires `TASK_CONFERENCE_TRANSFERRED → this.refreshTaskList`. But `handleTaskRemove` calls `taskToRemove.off(TASK_EVENTS.TASK_CONFERENCE_TRANSFERRED, this.handleConferenceEnded)` — wrong handler reference. This listener is **never actually removed**, causing a listener leak.
```
mkesavan13: More concerning part is that TASK_CONFERENCE_TRANSFERRED event doesn't have a UT at all today. We need a ticket for such gaps

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,486 @@
...
+
+### Key Insight: `refreshTaskList()` Elimination
+
+**Old:** 15+ events trigger `refreshTaskList()` → `cc.taskManager.getAllTasks()` → update store observables.
+
+**New:** SDK keeps `task.data` updated via state machine actions. The store can read `task.data` directly instead of re-fetching. `refreshTaskList()` should only be called for:
+- Initial load / hydration
+- Full page refresh recovery
+- `TASK_WRAPPEDUP` (task must be removed from list — may be replaceable with explicit list removal)
+
```
mkesavan13: I think the reason why we refreshTaskList is because the store doesn't listen to `task.data` directly to update the UI. While when we fetch the data manually from task object, it will show up. However, unless we set it again in store, it won't be observed and the UI won't re-render. How is that being tackled?

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,486 @@
...
+registerTaskEventListeners(task: ITask) {
+  const interactionId = task.data.interactionId;
+
+  // Create bound handlers that close over this task's interactionId.
+  // Stored in map so handleTaskRemove can .off() the exact same references.
+  const bound: Record<string, Function> = {
+    reject: (reason: string) => this.handleTaskReject(task, reason),
+    outdialFailed: (reason: string) => this.handleOutdialFailed(reason),
+    uiControlsUpdated: (uiControls: TaskUIControls) => {
+      this.fireTaskCallbacks(TASK_EVENTS.TASK_UI_CONTROLS_UPDATED, interactionId, uiControls);
```
mkesavan13: Even though it talks about `fireTaskCallbacks` a lot, it doesn't seem to have mentioned what it's definition would look like anywhere I believe. Can we check and include that?

mkesavan13 (changes requested):

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
```diff
@@ -0,0 +1,377 @@
+# Store Task Utils Migration
+
+## Summary
+
+The store's `task-utils.ts` contains 16 exported utility functions that inspect raw task data to derive state flags (consult status, hold status, conference state, participant info). Many of these become redundant when `task.uiControls` is the source of truth. This document maps which utils to keep, simplify, or remove.
+
+**Barrel export:** `store/src/index.ts` has `export * from './task-utils'` — all 16 functions are publicly exported via `@webex/cc-store`. Removing functions will cause compile errors in any downstream consumer still importing them.
```
mkesavan13: I don't think anyone will be using these utils. The only downstream we have today is Epic. So we can ask about these Utils usage in the space and eliminate them if they do not use it today. 

Exported doesn't mean used!

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
```diff
@@ -0,0 +1,377 @@
...
+
+---
+
+## Constants and Types to Delete
+
+| Delete | File | Reason |
+|--------|------|--------|
+| Local `TASK_EVENTS` enum | `store/src/store.types.ts` | SDK exports this — delete local copy (covered in detail in [store-event-wiring-migration.md](./store-event-wiring-migration.md)) |
+| `ConsultStatus` enum | `store/src/store.types.ts` | All consumers (`getConsultStatus`, `getControlsVisibility`) are being removed |
+| `TASK_STATE_CONSULT` | `store/src/constants.ts` | **Dual use:** (1) Task state → SDK `TaskState.CONSULT_INITIATING`. (2) **Media-type sentinel** in `findHoldStatus(task, mType, agentId)` — line 328 uses `mType === TASK_STATE_CONSULT` to identify the consult *leg*, not task state. **Delete ONLY AFTER rewriting `findHoldStatus`**; when rewriting, preserve a media-type constant (or use media-type enums) for the consult leg — do **not** replace that branch with `TaskState`. See ordering note below. |
```
mkesavan13: - TASK_STATE_CONSULT
- RELATIONSHIP_TYPE_CONSULT
- MEDIA_TYPE_CONSULT

All 3 of them resolve to the same string 'consult'

Is there anything we can do to tackle this?

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Doc updated: added a “Docs to update” note that points to CallControl ARCHITECTURE.md (consult sequence ~line 169) and says to update it when we migrate event names.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Doc updated: it now states the SDK doesn’t export TASK_EVENTS yet and points to migration-overview “SDK Pending Exports”; we keep the local enum until the SDK exports it.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Doc updated: added a “Migration / test plan” that when we wire TASK_CONSULT_END to handleConsultEnd, we’ll update store UTs to assert the new wiring and consult state reset.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Doc updated and tracked in Jira: added a “Test gap” note and linked CAI-7758 for adding UTs for this and similar gaps.
https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7758

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Doc updated: added subsection “Why we can remove most refreshTaskList()” explaining SDK updates the same task reference, widgets re-render via fireTaskCallbacks/setTaskCallback, and we only keep refreshTaskList where the list itself changes.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Doc updated: added “fireTaskCallbacks — Definition” with purpose, signature, and location (storeEventsWrapper.ts).

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
```diff
@@ -0,0 +1,377 @@
...
+
+| Keep | File | Reason |
+|------|------|--------|
+| `RELATIONSHIP_TYPE_CONSULT` | `store/src/constants.ts` | Still used by `findMediaResourceId` (KEEP) |
+| `MEDIA_TYPE_CONSULT` | `store/src/constants.ts` | Still used by `findMediaResourceId` (KEEP) |
+| `AGENT` | `store/src/constants.ts` | Used by `getConferenceParticipants` (KEEP) for participant filtering |
+| `CUSTOMER` | `store/src/constants.ts` | Used by `EXCLUDED_PARTICIPANT_TYPES` |
+| `SUPERVISOR` | `store/src/constants.ts` | Used by `EXCLUDED_PARTICIPANT_TYPES` |
+| `VVA` | `store/src/constants.ts` | Used by `EXCLUDED_PARTICIPANT_TYPES` |
+| `EXCLUDED_PARTICIPANT_TYPES` | `store/src/constants.ts` | Used by `getConferenceParticipants` (KEEP) for participant filtering |
```
mkesavan13: Why do some say (KEEP) and a few doesn't? All of these are supposed to be kept. Is my understanding correct?

Some of them might change based on my comment above about 3 constants resolving to same string 'consult'

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,509 @@
...
+| 14 | `TASK_OFFER_CONTACT` | Fire callback only | **Simplify + rename** | Was `AGENT_OFFER_CONTACT`. Remove `refreshTaskList()`. |
+| 15 | `TASK_OFFER_CONSULT` | Fire callback only | **Simplify** | Remove `refreshTaskList()`. |
+| 16 | `TASK_PARTICIPANT_JOINED` | `handleConferenceStarted` | **Simplify** | Keep consult state reset (`isQueueConsultInProgress`, `currentConsultQueueId`, `consultStartTimeStamp`). Remove `refreshTaskList()`. Fire callback. |
+| 17 | `TASK_CONFERENCE_STARTED` | `handleConferenceStarted` | **Simplify** | Same as #16 |
+| 18 | `TASK_CONFERENCE_ENDED` | `handleConferenceEnded` | **Simplify** | Remove `refreshTaskList()`. Fire callback. |
+| 19 | `TASK_PARTICIPANT_LEFT` | `handleConferenceEnded` | **Simplify** | Same as #18 |
+| 20 | `TASK_HOLD` | Fire callback only | **Simplify** | Remove `refreshTaskList()`. |
+| 21 | `TASK_RESUME` | Fire callback only | **Simplify** | Remove `refreshTaskList()`. |
+| 22 | `TASK_RECORDING_PAUSED` | Fire callback only | **Simplify + rename** | Was `CONTACT_RECORDING_PAUSED`. |
```
rsarika: will be removed right?

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,509 @@
...
+| 23 | `TASK_RECORDING_RESUMED` | Fire callback only | **Simplify + rename** | Was `CONTACT_RECORDING_RESUMED`. |
```
rsarika: will be removed right?

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
rsarika: but still refreshTaskList needed right?

rsarika: we can enahance refreshTaskList to take task as param and update only given task

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,509 @@
...
+- Initial load / hydration
+- Full page refresh recovery
+- `TASK_WRAPPEDUP` (task must be removed from list — may be replaceable with explicit list removal)
+
+#### Why we can remove most `refreshTaskList()`
+
+1. **SDK keeps the same task reference up to date.** The state machine updates `task.data` (and `task.uiControls`) on the **same** `ITask` reference already held in the store's `taskList`. No re-fetch is needed for in-place updates.
+2. **Widget re-renders are driven by callbacks.** Widgets that registered via `setTaskCallback(event, cb, taskId)` are notified when the store calls `fireTaskCallbacks(event, interactionId, payload)`. Those callbacks cause the widget to re-run and re-render with the updated task from the store.
+3. **Re-fetch is only needed when the list itself changes.** We keep `refreshTaskList()` only where the **list** must change: e.g. initial load, full refresh, or `TASK_WRAPPEDUP` (task removed from the list). For all other events, the existing task reference is already updated by the SDK, and `fireTaskCallbacks` triggers UI updates.
```
rsarika: this true w.r.t sdk perspective. but widgets still need to get that latest updated data. decide on how? whether with refresh task list or not

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,509 @@
...
+The widget's local `TASK_EVENTS` enum (in `store/src/store.types.ts`) uses CC-level naming that differs from the SDK's task-level naming.
+
+### 5 Renamed Events
+
+| Old (Widget) | Old Value | New (SDK) | New Value |
+|---|---|---|---|
+| `AGENT_WRAPPEDUP` | `'AgentWrappedUp'` | `TASK_WRAPPEDUP` | `'task:wrappedup'` |
+| `AGENT_CONSULT_CREATED` | `'AgentConsultCreated'` | `TASK_CONSULT_CREATED` | `'task:consultCreated'` |
+| `AGENT_OFFER_CONTACT` | `'AgentOfferContact'` | `TASK_OFFER_CONTACT` | `'task:offerContact'` |
+| `CONTACT_RECORDING_PAUSED` | `'ContactRecordingPaused'` | `TASK_RECORDING_PAUSED` | `'task:recordingPaused'` |
```
Kesari3008, line 118: These recording event names are redundant. I see in TASK_EVENTS, we already have task:recordingPaused and task:recordingResumed added. So we can just remove these names and this is not again something that needs to be done as aprt of task-refactor I believe

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,509 @@
...
+| `CONTACT_RECORDING_RESUMED` | `'ContactRecordingResumed'` | `TASK_RECORDING_RESUMED` | `'task:recordingResumed'` |
+
+### 4 Store-Only Enum Members (no SDK equivalent — delete)
+
+| Widget Enum Member | Value | Note |
+|---|---|---|
+| `TASK_UNHOLD` | `'task:unhold'` | SDK uses `TASK_RESUME` instead |
```
Kesari3008: Attach the actual event string here what TASK_RESUME and TASK_HOLD is

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -173,6 +179,23 @@ Many events that currently trigger `refreshTaskList()` will no longer need it be
...
+
+### fireTaskCallbacks — Definition
+
+**Purpose:** Invoke all callbacks that were registered for a given task event via `setTaskCallback(event, cb, taskId)` (or equivalent), so widgets can re-render or react when that event occurs.
+
+**Signature (conceptual):**  
+`fireTaskCallbacks(event: TASK_EVENTS, interactionId: string, payload?: unknown): void`
+
+**Where it lives:** In the store-events layer — `packages/contact-center/store/src/storeEventsWrapper.ts`. After the migration, "After" handlers call `fireTaskCallbacks(...)` instead of (or in addition to) `refreshTaskList()` for events that only need to notify widgets. Implementation pattern: look up callbacks by event and optional `taskId`/`interactionId`, then invoke each with the payload.
+
```
mkesavan13: This again doesn't give details of how the callbacks or fireTaskCalback will update the task data back into the store. That implementation information is required

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
```diff
@@ -0,0 +1,377 @@
...
+| `TASK_STATE_CONSULT_COMPLETED` | `store/src/constants.ts` | SDK handles via context — **same ordering constraint** |
+| `INTERACTION_STATE_WRAPUP` | `store/src/constants.ts` | SDK `TaskState.WRAPPING_UP` — **delete ONLY AFTER rewriting `getTaskStatus`** (see ordering note below) |
+| `INTERACTION_STATE_POST_CALL` | `store/src/constants.ts` | SDK `TaskState.POST_CALL` — **same ordering constraint** |
+| `INTERACTION_STATE_CONNECTED` | `store/src/constants.ts` | SDK `TaskState.CONNECTED` — **same ordering constraint** |
+| `INTERACTION_STATE_CONFERENCE` | `store/src/constants.ts` | SDK `TaskState.CONFERENCING` — **same ordering constraint** |
+| `CONSULT_STATE_INITIATED` | `store/src/constants.ts` | SDK handles via context — **delete ONLY AFTER rewriting `getConsultMPCState`** (see ordering note below) |
+| `CONSULT_STATE_COMPLETED` | `store/src/constants.ts` | SDK handles via context — **same ordering constraint** |
+| `CONSULT_STATE_CONFERENCING` | `store/src/constants.ts` | SDK handles via context — **same ordering constraint** |
+
+## Constants to Keep
```
rsarika: we should remove all these, the task object should be source of truth. we should not have these kind of utils in widgets

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
```diff
@@ -0,0 +1,377 @@
...
+|------|------|--------|
+| `RELATIONSHIP_TYPE_CONSULT` | `store/src/constants.ts` | Still used by `findMediaResourceId` (KEEP) |
+| `MEDIA_TYPE_CONSULT` | `store/src/constants.ts` | Still used by `findMediaResourceId` (KEEP) |
+| `AGENT` | `store/src/constants.ts` | Used by `getConferenceParticipants` (KEEP) for participant filtering |
+| `CUSTOMER` | `store/src/constants.ts` | Used by `EXCLUDED_PARTICIPANT_TYPES` |
+| `SUPERVISOR` | `store/src/constants.ts` | Used by `EXCLUDED_PARTICIPANT_TYPES` |
+| `VVA` | `store/src/constants.ts` | Used by `EXCLUDED_PARTICIPANT_TYPES` |
+| `EXCLUDED_PARTICIPANT_TYPES` | `store/src/constants.ts` | Used by `getConferenceParticipants` (KEEP) for participant filtering |
+
+## Ordering Constraint: Consult State Constants
```
rsarika: below also should remove , the task object should be source of truth. we should not have these kind of utils in widgets

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
```diff
@@ -0,0 +1,377 @@
...
+
+`findHoldStatus` (KEEP) depends on `TASK_STATE_CONSULT`, `TASK_STATE_CONSULTING`, and `TASK_STATE_CONSULT_COMPLETED` via:
+- **Direct usage:** Line 328 — `mType === TASK_STATE_CONSULT`
+- **Via `isConsultOnHoldMPC`:** Line 303 — `[TASK_STATE_CONSULT, TASK_STATE_CONSULTING].includes(getConsultMPCState(...))`
+- **Via `getConsultMPCState`:** Line 321 — `[TASK_STATE_CONSULT_COMPLETED].includes(getConsultMPCState(...))`
+
+**Do NOT delete these 3 constants until `findHoldStatus` and `isConsultOnHoldMPC` are rewritten** to use SDK `TaskState` equivalents. Deleting them first will break compilation. When rewriting `findHoldStatus`, note that `TASK_STATE_CONSULT` is used there as a **media-type** sentinel (`mType === TASK_STATE_CONSULT`); preserve a media-type constant or media-type enum for the consult leg — do not replace that comparison with `TaskState`.
```
rsarika: we should remove these as now task is source of truth

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,509 @@
...
+The store's `storeEventsWrapper.ts` currently registers 27 individual task event handlers via `registerTaskEventListeners()` that manually call `refreshTaskList()` and update observables. A companion method `handleTaskRemove()` unregisters them. With the state machine, the SDK handles state transitions internally. Many event handlers can be simplified or removed, and the new `task:ui-controls-updated` event replaces manual state derivation.
+
+---
+
+## Event Names — Renamed and Deleted
+
+The widget's local `TASK_EVENTS` enum (in `store/src/store.types.ts`) uses CC-level naming that differs from the SDK's task-level naming.
+
+### 5 Renamed Events
```
Kesari3008: Some of the details here don't seem correct completely. Here is the complete list of what events are being emitted from SDK, based on this information update the events that need change:

From Task.ts
TASK_EVENTS.TASK_INCOMING
TASK_EVENTS.TASK_HYDRATE
TASK_EVENTS.TASK_OFFER_CONTACT
TASK_EVENTS.TASK_ASSIGNED
TASK_EVENTS.TASK_END
TASK_EVENTS.TASK_OFFER_CONSULT
TASK_EVENTS.TASK_CONSULT_CREATED
TASK_EVENTS.TASK_CONSULT_ACCEPTED
TASK_EVENTS.TASK_CONSULT_END
TASK_EVENTS.TASK_CONSULT_QUEUE_CANCELLED
TASK_EVENTS.TASK_CONSULT_QUEUE_FAILED
TASK_EVENTS.TASK_WRAPPEDUP

From voice/Voice.ts
TASK_EVENTS.TASK_HOLD
TASK_EVENTS.TASK_RESUME
TASK_EVENTS.TASK_RECORDING_STARTED
TASK_EVENTS.TASK_RECORDING_PAUSED
TASK_EVENTS.TASK_RECORDING_PAUSE_FAILED
TASK_EVENTS.TASK_RECORDING_RESUMED
TASK_EVENTS.TASK_RECORDING_RESUME_FAILED
TASK_EVENTS.TASK_PARTICIPANT_JOINED
TASK_EVENTS.TASK_PARTICIPANT_LEFT
TASK_EVENTS.TASK_CONFERENCE_STARTED
TASK_EVENTS.TASK_CONFERENCE_ENDED
TASK_EVENTS.TASK_CONFERENCE_FAILED
TASK_EVENTS.TASK_EXIT_CONFERENCE
TASK_EVENTS.TASK_TRANSFER_CONFERENCE
TASK_EVENTS.TASK_SWITCH_CALL
TASK_EVENTS.TASK_CONFERENCE_TRANSFER_FAILED
TASK_EVENTS.TASK_OUTDIAL_FAILED

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
```diff
@@ -0,0 +1,377 @@
...
+| 9 | `getConferenceParticipantsCount(task)` | Participant count | `task/src/Utils/task-util.ts` (`getControlsVisibility`) |
+| 10 | `getIsCustomerInCall(task)` | Whether customer is connected | `task/src/Utils/task-util.ts` (`getControlsVisibility`) |
+| 11 | `getIsConsultInProgress(task)` | Whether consult is active | `task/src/Utils/task-util.ts` (`getControlsVisibility`) |
+| 12 | `isInteractionOnHold(task)` | Whether any media is held | Timer utils |
+| 13 | `setmTypeForEPDN(task, mType)` | Media type for EP-DN agents | CallControl hook |
+| 14 | `findMediaResourceId(task, mType)` | Media resource ID lookup | CallControl hook (switch calls) |
+| 15 | `findHoldStatus(task, mType, agentId)` | Hold status by media type | `task/src/Utils/task-util.ts` (`getControlsVisibility`), `getTaskStatus()` (via `getConsultMPCState` chain) |
+| 16 | `findHoldTimestamp(task, mType)` | Hold timestamp for timers | Timer utils (store version takes `ITask`; see dual-signature note below) |
+
+### Important: `findHoldTimestamp` Dual Signatures
```
rsarika: not needed anymore

mkesavan13 (changes requested):

rsarika (changes requested):

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,509 @@
...
+| `TASK_CONSULT_QUEUE_FAILED` | `'task:consultQueueFailed'` | Evaluate if widget needs to handle |
+| `TASK_RECORDING_STARTED` | `'task:recordingStarted'` | Evaluate for recording indicator |
+| `TASK_RECORDING_PAUSE_FAILED` | `'task:recordingPauseFailed'` | Evaluate for error handling |
+| `TASK_RECORDING_RESUME_FAILED` | `'task:recordingResumeFailed'` | Evaluate for error handling |
+| `TASK_EXIT_CONFERENCE` | `'task:exitConference'` | Evaluate for conference flow |
+| `TASK_TRANSFER_CONFERENCE` | `'task:transferConference'` | Evaluate for conference flow |
+| `TASK_CLEANUP` | `'task:cleanup'` | SDK internal — likely no widget action |
+
+**Action:** Delete the local `TASK_EVENTS` enum from `store/src/store.types.ts` and import from SDK once the SDK exports it. **As of this migration, the SDK does not yet export `TASK_EVENTS` from its package entry point** (or status is TBD). For current status and the plan to replace the local enum once the SDK exports it, see [migration-overview.md](./migration-overview.md) → "SDK Pending Exports" (or equivalent section). Until then, keep the local enum and align event string values with the SDK where needed.
```
Kesari3008: We are exporting TASK_EVENTS from the SDK. Check index.ts file: /Users/pkesari/Desktop/WorkProjects/WebRTCFrontEnd/wxcc-sdk/webex-js-sdk/packages/@webex/contact-center/src/index.ts

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,509 @@
...
+| 20 | `TASK_PARTICIPANT_JOINED` | `handleConferenceStarted` | Reset `isQueueConsultInProgress`, `currentConsultQueueId`, `consultStartTimeStamp` + `refreshTaskList()` |
+| 21 | `TASK_PARTICIPANT_LEFT` | `handleConferenceEnded` | `refreshTaskList()` |
+| 22 | `TASK_PARTICIPANT_LEFT_FAILED` | `refreshTaskList` | Re-fetch all tasks |
+| 23 | `TASK_CONFERENCE_STARTED` | `handleConferenceStarted` | (same as #20) |
+| 24 | `TASK_CONFERENCE_TRANSFERRED` | `refreshTaskList` | Re-fetch all tasks |
+| 25 | `TASK_CONFERENCE_TRANSFER_FAILED` | `refreshTaskList` | Re-fetch all tasks |
+| 26 | `TASK_POST_CALL_ACTIVITY` | `refreshTaskList` | Re-fetch all tasks |
+| 27 | `TASK_MEDIA` | `handleTaskMedia` | Browser-only: `setCallControlAudio(new MediaStream([track]))` |
+
+### Pre-existing Bugs in Old Code
```
Kesari3008: Would this be relevant for task-refactor specific migrations ?

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,509 @@
...
+3. `task.uiControls` is recomputed after every state transition
+4. `task:ui-controls-updated` is emitted when controls change
+
+### Definitive New Event Registration
+
+Many events that currently trigger `refreshTaskList()` will no longer need it because `task.data` is kept in sync by the SDK. Below is the single authoritative table for all event handler changes:
+
+| # | Event | New Handler | Change | Detail |
+|---|-------|-------------|--------|--------|
+| 1 | `TASK_END` | `handleTaskEnd` | **Keep** | Remove from task list, clear current task |
```
Kesari3008: What does Change column refer to ? Does Keep means no change is required for the handler specified ? I don't see much usecase of this column when we have Detail column where we can mention in bullet points change to be done

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,509 @@
...
+| 2 | `TASK_ASSIGNED` | `handleTaskAssigned` | **Keep** | Update task list, set current task |
+| 3 | `TASK_REJECT` | `handleTaskReject` | **Keep** | Remove from task list |
+| 4 | `TASK_OUTDIAL_FAILED` | `handleOutdialFailed` | **Keep** | Remove from task list |
+| 5 | `TASK_MEDIA` | `handleTaskMedia` | **Keep** | Browser-only WebRTC setup (conditional registration) |
+| 6 | `TASK_UI_CONTROLS_UPDATED` | `bound.uiControlsUpdated` (per-task; see Pattern 1 & 3) | **Add new** | Fire callbacks to trigger widget re-renders. Do **not** use a class-level handler — it would resolve the wrong `interactionId` in multi-task scenarios. |
```
Kesari3008: New event addition and handler should be separate from sections where are modifying existing change.

Order this table in this manner, start with the list of handlers which do not need change, then handlers which require change and at last new additions can be added

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
akulakum: Doc updated: added a pre-migration step to confirm downstream usage (e.g. Epic) before removing utils — "exported doesn’t mean used."

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
akulakum: Doc updated: added “Consult string alias” in Constants to Keep — all three are documented as the same string `'consult'`; we can consider one constant or document the alias; deletion/rewrite ordering is unchanged.

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
akulakum: Yes — all entries in the “Constants to Keep” table are kept. Doc updated so the Reason column is consistent and the consult string alias (TASK_STATE_CONSULT / RELATIONSHIP_TYPE_CONSULT / MEDIA_TYPE_CONSULT) is called out; any consolidation would follow from that.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Doc updated: added an implementation note that fireTaskCallbacks does **not** write task data back. The SDK mutates the same task reference in place; the store holds that reference in taskList; when we call fireTaskCallbacks we only invoke widget callbacks, and widgets get the latest data by re-reading the same task from the store. So no “update back into the store” step is needed for in-place task updates.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Only the **call** to `refreshTaskList()` is removed from those handlers. The **handlers themselves are kept** — they still fire callbacks so widgets can re-render. The doc table and “When refreshTaskList() is still required” section spell this out.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Same as above: we remove only `refreshTaskList()` from the handler; we keep the handler and fire callbacks. Remove = remove refresh call, not the handler.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Yes. refreshTaskList() is still needed for: initial load/hydration, full refresh recovery, and TASK_WRAPPEDUP (task must be removed from the list). Doc has a “When refreshTaskList() is still required” subsection. For other events we only remove the refresh call and keep firing callbacks.

Noted. Doc updated with a “Future consideration” that we could add an optional single-task refresh, e.g. `refreshTaskList(task?)`, to update only the given task.

Doc clarifies: widgets get the latest data by re-reading the **same** task reference from the store (SDK mutates it in place). We don’t need refreshTaskList for that — we only need it when the **list** changes (init, full refresh, TASK_WRAPPEDUP). For other events, fireTaskCallbacks triggers re-renders and widgets read the updated task from the store.

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
akulakum: Agreed — task is the source of truth and we’re removing store task-utils. Doc states that. The “Constants to Delete” table and ordering constraint say to delete these **only after** findHoldStatus / related helpers are rewritten (or SDK/task exposes equivalent). Until then we keep the constants so we don’t break compilation or hold derivation.

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
akulakum: Same as above: task as source of truth is the end state. We remove the store utils and constants once we have SDK/task equivalents. The doc’s ordering constraint says do not delete those constants until findHoldStatus (and related) are rewritten; then we can remove them.

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
```diff
@@ -0,0 +1,389 @@
...
+
+## Gotcha: `TaskState.CONSULT_INITIATING` vs `CONSULTING`
+
+The SDK has `CONSULT_INITIATING` (consult requested, async in-progress) and `CONSULTING` (consult accepted, actively consulting) as distinct states. The old widget constant `TASK_STATE_CONSULT` ('consult') maps to `CONSULT_INITIATING`, NOT `CONSULTING`. Do not collapse these when updating `getTaskStatus()` or any consult timer logic.
+
+## Decision: `findHoldStatus` and `findHoldTimestamp` retained for now
+
+Reviewers may suggest removing these because "task is source of truth." They are **kept** in this migration because: (1) SDK `task.uiControls.hold.isEnabled` indicates whether the hold *action* is available, not whether a given leg is *currently held* (e.g. during consult, main call can be held while consult is active). (2) Timers and UI need per-leg hold state and hold timestamp, which are read from `task.data.interaction.media`; the store helpers centralize that derivation. Once the SDK or task layer exposes equivalent hold state (or widgets derive it in a single place from `task.data` only), these helpers can be removed and the "task as source of truth" end state is fully achieved.
```
mkesavan13: > (1) SDK `task.uiControls.hold.isEnabled` indicates whether the hold *action* is available, not whether a given leg is *currently held* (e.g. during consult, main call can be held while consult is active).

This statement is incorrect. The UI Controls in the task object (`task.uiControls`) is recomputed every time the task.data changes. So, it should be considered for anything and everything in the UI

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,523 @@
...
+
+**`refreshTaskList()` remains required** for these cases only:
+
+1. **Initial load / hydration** — populate the store's task list when the app or CC session loads.
+2. **Full page refresh recovery** — re-sync the list after a full page reload.
+3. **`TASK_WRAPPEDUP`** — the task must be removed from the list after wrap-up completion; today that is done by calling `refreshTaskList()` (or may be replaced later with explicit list removal).
+
+For all other events (e.g. `TASK_HOLD`, `TASK_RESUME`, `TASK_RECORDING_PAUSED`, consult/conference lifecycle), **do not** call `refreshTaskList()`. The handler stays; it only fires callbacks so widgets re-render.
+
+#### Why we can remove most `refreshTaskList()` calls
```
Kesari3008: We don't really need to explain why we need to do something but just mentioned what needs to be changes in bullet points and that should be enough

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,523 @@
...
+
+**Where it lives:** In the store-events layer — `packages/contact-center/store/src/storeEventsWrapper.ts`. After the migration, "After" handlers call `fireTaskCallbacks(...)` instead of (or in addition to) `refreshTaskList()` for events that only need to notify widgets. Implementation pattern: look up callbacks by event and optional `taskId`/`interactionId`, then invoke each with the payload.
+
+**Implementation note — task data flow (no write-back):**  
+`fireTaskCallbacks` **does not** write or update task data back into the store. The flow is: (1) The SDK mutates the **same** `ITask` reference already held in the store's `taskList` (updates `task.data` and `task.uiControls` in place). (2) The store calls `fireTaskCallbacks(event, interactionId, payload)`, which only **invokes** the registered widget callbacks. (3) Widgets then **read** from the store (e.g. `store.taskList[interactionId]` or `store.currentTask`) and re-render with the updated task. So widgets get the latest data by re-reading that same reference after the callback; there is no separate "update task data back into the store" step.
```
mkesavan13: How would the widgets read the data update? In the store, we have marked the `cc` object to be not observed in the constructor?

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,523 @@
...
+// task.data and task.uiControls already reflect conference state.
+// Keep consult state reset in handleConferenceStarted; remove refreshTaskList() from both.
+//
+// Both eventType and interactionId are passed by the bound handlers in Pattern 1.
+// This avoids: (a) dual callback firing, (b) unresolved interactionId in class scope,
+// and (c) inline arrows that can't be detached by handleTaskRemove.
+
+handleConferenceStarted = (eventType: TASK_EVENTS, interactionId: string) => {
+  runInAction(() => {
+    this.setIsQueueConsultInProgress(false);
```
Kesari3008: Would we still need these after the migration ? What does these setters help us with today ?

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,523 @@
...
+  }
+
+  // Reset store state
+  // ... existing currentTask/taskList cleanup logic
+};
+```
+
+---
+
+## Files to Modify
```
Kesari3008: This section should be placed right after summary

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,523 @@
...
+| 3 | `AGENT_OFFER_CONTACT` | `refreshTaskList` | Re-fetch all tasks |
+| 4 | `AGENT_CONSULT_CREATED` | `handleConsultCreated` | `refreshTaskList()` + `setConsultStartTimeStamp(Date.now())` |
+| 5 | `TASK_CONSULT_QUEUE_CANCELLED` | `handleConsultQueueCancelled` | Reset `isQueueConsultInProgress`, `currentConsultQueueId`, `consultStartTimeStamp` + `refreshTaskList()` |
+| 6 | `TASK_REJECT` | `handleTaskReject` | Remove task, fire callbacks |
+| 7 | `TASK_OUTDIAL_FAILED` | `handleOutdialFailed` | Remove task, fire callbacks |
+| 8 | `AGENT_WRAPPEDUP` | `refreshTaskList` | Re-fetch all tasks |
+| 9 | `TASK_CONSULTING` | `handleConsulting` | `refreshTaskList()` + `setConsultStartTimeStamp(Date.now())` |
+| 10 | `TASK_CONSULT_ACCEPTED` | `handleConsultAccepted` | `refreshTaskList()` + `setConsultStartTimeStamp(Date.now())` + set ENGAGED state + **registers `TASK_MEDIA` listener on consult task (browser only)** |
+| 11 | `TASK_OFFER_CONSULT` | `handleConsultOffer` | `refreshTaskList()` |
+| 12 | `TASK_AUTO_ANSWERED` | `handleAutoAnswer` | `setIsDeclineButtonEnabled(true)` + `refreshTaskList()` |
```
Kesari3008, line 295: Why do we need isDeclineButtonEnabled property and getter and setters for this ? UIControls should get used in place of this and we are not adding this point in this migrationdocs that this needs to be replaced with uiCOntrols.decline.isEnabled

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
```diff
@@ -0,0 +1,523 @@
...
+3. **Re-fetch is only needed when the list itself changes.** We keep `refreshTaskList()` only where the **list** must change: initial load, full refresh, or `TASK_WRAPPEDUP` (task removed from the list). For all other events, the existing task reference is already updated by the SDK, and `fireTaskCallbacks` triggers UI updates.
+
+#### Future consideration: optional single-task refresh
+
+As an enhancement, `refreshTaskList()` could be extended to accept an optional task (or `interactionId`) and update only that task in the store instead of re-fetching the full list (e.g. for wrap-up or other single-task updates). This is not required for the current migration.
+
+---
+
+### fireTaskCallbacks — Definition
```
Kesari3008: This whole approach is not task-refactor migration related ? So why are we coming up with as part of this migration doc ? What is the problem that we are trying to solve here ? And if we do wanna address some of the refactoring in widgets to solve some existing problems then it would be good to separate the sections. Modifcations based on task-refactor changes and modifications for Widgets imporvements and fixing exiisting issues. Also I feel this is unnecessary, we are try to over complicate a simple approach of switching on and witching off the listeners with their callbacks associated

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
```diff
@@ -0,0 +1,389 @@
...
+
+---
+
+## Files to Modify
+
+| File | Action |
+|------|--------|
+| `store/src/task-utils.ts` | Remove 5 functions, keep 7 (update `getTaskStatus` to use `task.uiControls`), review 4 |
+| `store/src/store.types.ts` | Delete `ConsultStatus` enum (all consumers removed) |
+| `store/src/constants.ts` | Delete 9 task/interaction/consult state constants; keep 7 participant/media constants |
```
Kesari3008: This table says delete 9 constants, but the "After" getTaskStatus code still uses them as return values. Clarify that these deletions are deferred until a subsequent step where getTaskStatus is further rewritten to use SDK TaskState values.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Doc updated: added a short “How widgets see updates when the store does not observe the cc object” note. Because the store doesn’t observe cc (or task internals), MobX doesn’t re-run when the SDK mutates the task in place. Re-renders are callback-driven: when the store calls fireTaskCallbacks, the widget’s callback runs (e.g. setState or re-read), then the widget reads store.currentTask / store.taskList[interactionId] — the same task reference the SDK has already updated — so it gets the latest task.data and task.uiControls. So the callback triggers the re-render and read; we don’t rely on an observable on task internals

Kesari3008 (dismissed):

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Fixed. Moved recording events out of the "Renamed Events" section into a new "Pre-existing Name Mismatches" section — these are not task-refactor renames, just incorrect names in the widget's local enum that already differed from the SDK.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Added SDK string values — TASK_RESUME ('task:resume') and TASK_HOLD ('task:hold') — in the notes for TASK_UNHOLD and TASK_PAUSE.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Updated the SDK Event Reference section with the complete list from Task.ts and Voice.ts, including string values. Also refined "5 Renamed Events" to "3 Renamed Events (task-refactor specific)" and added a separate "Pre-existing Name Mismatches" section for recording events.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Doc now confirms SDK exports TASK_EVENTS from index.ts and instructs to delete the local enum and import from @webex/contact-center, with a fallback note for older SDK versions.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: TASK_MEDIA is marked as "Unchanged — not task-refactor related" in the definitive table (row 5 in "No change required" section). It's browser-only WebRTC setup and untouched by this migration.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Removed the "Change" column (merged into Detail).

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Restructured the table into three sub-sections: "No change required (22 events)", "Changes required (7 events)", and "New additions (1 event)".

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Added a "Store Consult State Observables (future removal)" section explaining these are retained because widget consult timer logic reads consultStartTimeStamp directly. End state: remove once widgets derive consult timing from task.data

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Moved "Files to Modify" to immediately after the Summary section.

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Added a dedicated "Migration: isDeclineButtonEnabled → task.uiControls.decline.isEnabled" section with consumers, migration steps, and updated handleAutoAnswer Before/After to remove setIsDeclineButtonEnabled(true)

packages/contact-center/ai-docs/migration/store-event-wiring-migration.md
akulakum: Removed the fireTaskCallbacks definition, bound-handler map pattern, and related complexity. Simplified the approach: refreshTaskList() is kept in all existing handlers (not removed). The doc now clearly separates task-refactor changes from future widget optimizations. The refreshTaskList() removal is documented as a future widget improvement, not part of this migration.

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
akulakum: The store/src/constants.ts row now reads: "Phased deletion: Delete 9 constants only after dependent functions (getTaskStatus, findHoldStatus, getConsultMPCState) are rewritten to use SDK TaskState."

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
```diff
@@ -0,0 +1,389 @@
...
+| `INTERACTION_STATE_POST_CALL` | `store/src/constants.ts` | SDK `TaskState.POST_CALL` — **same ordering constraint** |
+| `INTERACTION_STATE_CONNECTED` | `store/src/constants.ts` | SDK `TaskState.CONNECTED` — **same ordering constraint** |
+| `INTERACTION_STATE_CONFERENCE` | `store/src/constants.ts` | SDK `TaskState.CONFERENCING` — **same ordering constraint** |
+| `CONSULT_STATE_INITIATED` | `store/src/constants.ts` | SDK handles via context — **delete ONLY AFTER rewriting `getConsultMPCState`** (see ordering note below) |
+| `CONSULT_STATE_COMPLETED` | `store/src/constants.ts` | SDK handles via context — **same ordering constraint** |
+| `CONSULT_STATE_CONFERENCING` | `store/src/constants.ts` | SDK handles via context — **same ordering constraint** |
+
+## Constants to Keep
+
+**All entries in this table are kept** (no deletions). Use them until the corresponding helpers are rewritten or removed per the ordering constraints below.
```
rsarika: we still need them to be removed right?

rsarika (approved):

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
akulakum: Removed

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
```diff
@@ -0,0 +1,178 @@
...
+
+### What gets deleted from `task-util.ts`
+
+`getControlsVisibility` + all 22 `get*ButtonVisibility` functions:
+`getAcceptButtonVisibility`, `getDeclineButtonVisibility`, `getEndButtonVisibility`, `getMuteUnmuteButtonVisibility`, `getHoldResumeButtonVisibility`, `getPauseResumeRecordingButtonVisibility`, `getRecordingIndicatorVisibility`, `getTransferButtonVisibility`, `getConferenceButtonVisibility`, `getExitConferenceButtonVisibility`, `getMergeConferenceButtonVisibility`, `getConsultButtonVisibility`, `getEndConsultButtonVisibility`, `getConsultTransferButtonVisibility`, `getMergeConferenceConsultButtonVisibility`, `getConsultTransferConsultButtonVisibility`, `getMuteUnmuteConsultButtonVisibility`, `getSwitchToMainCallButtonVisibility`, `getSwitchToConsultButtonVisibility`, `getWrapupButtonVisibility`
+
+### What replaces it
+
+```typescript
+const controls = currentTask?.uiControls ?? getDefaultUIControls();
```
mkesavan13, line 141: I do not see any definition for `getDefaultUIControls` anywhere.

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
```diff
@@ -0,0 +1,178 @@
...
+| `featureFlags.webRtcEnabled` (recording gate) | `config.isRecordingEnabled` |
+| `conferenceEnabled` | SDK computes conference/mergeToConference/exitConference visibility based on task state and config |
+
+Since `task.uiControls` already reflects these gates, the widget layer can **remove** the `featureFlags`, `conferenceEnabled`, and `deviceType` props — no widget-side overlay is needed.
+
+```typescript
+const controls = currentTask?.uiControls ?? getDefaultUIControls();
+```
+
+> **Note:** There is no `applyFeatureGates` function in the SDK or widgets codebase. The old feature-flag logic was inlined in `getControlsVisibility` (lines 421–525 of `task-util.ts`). It is deleted along with `getControlsVisibility`.
```
mkesavan13: Why does it talk about `applyFeatureGates` out of the blue?

mkesavan13 (approved):
Just a couple of comments. Rest looks good to me.

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
akulakum: Removed — it was a leftover note from an earlier review round; the SDK handles feature-flag gating internally via UIControlConfig, so no applyFeatureGates reference is needed

packages/contact-center/ai-docs/migration/store-task-utils-migration.md
akulakum: It's exported by the SDK from uiControlsComputer.ts — returns all 17 controls as { isVisible: false, isEnabled: false }; added the import and definition to the doc.

PR #649, opened 2026-03-11
chore: add spec-drift detection commands and pre-commit hook

ciscoRankush:

Adding automated SDD (Spec-Driven Development) documentation drift detection tooling to ensure ai-docs stay in sync with source code.

- **`.claude/commands/spec-drift.md`** — Full scan command that auto-discovers all ai-docs folders, spawns parallel checker agents, and validates 7 drift categories (file tree, method/API, types, events, architecture, links, code examples)
- **`.claude/commands/spec-drift-changed.md`** — Incremental scan that only validates ai-docs affected by staged/unstaged file changes, with verification marker creation
- **`.claude/hooks/check-ai-docs-drift.sh`** — Pre-commit blocking hook that prevents committing when ai-docs drift is unverified

### Change Type

- [x] Tooling change

- Verified spec-drift command discovers all ai-docs folders and produces a categorized report
- Verified spec-drift-changed detects changed files and maps them to their ai-docs
- Verified pre-commit hook blocks commits with unverified ai-docs changes (exit code 2)
- Verified hook passes after spec-drift-changed creates verification marker
- Verified hook is transparent for commits without ai-docs-related changes

## The GAI Coding Policy And Copyright Annotation Best Practices ##

- [x] GAI was used to create a draft that was subsequently customized or modified
- [x] Tool used for AI assistance (GitHub Copilot / Other - specify)
  - [x] Other - Claude Code (Anthropic)
- [x] This PR is related to
  - [x] Automation

### I certified that

- [x] I have read and followed contributing guidelines
- [x] I discussed changes with code owners prior to submitting this pull request
- [x] I have updated the documentation accordingly

### DIFF
--- a/.claude/commands/spec-drift-changed.md
+++ b/.claude/commands/spec-drift-changed.md
@@ -37,11 +37,10 @@ packages/contact-center/task/src/widgets/CallControl/CallControl.tsx → package
 packages/contact-center/cc-components/src/SomeComponent.tsx        → packages/contact-center/cc-components/ai-docs/
 ```
 
-For task widget source files, map to the widget-specific ai-docs:
-- `packages/contact-center/task/src/widgets/CallControl/*` → `packages/contact-center/task/ai-docs/widgets/CallControl/`
-- `packages/contact-center/task/src/widgets/IncomingTask/*` → `packages/contact-center/task/ai-docs/widgets/IncomingTask/`
-- `packages/contact-center/task/src/widgets/OutdialCall/*` → `packages/contact-center/task/ai-docs/widgets/OutdialCall/`
-- `packages/contact-center/task/src/widgets/TaskList/*` → `packages/contact-center/task/ai-docs/widgets/TaskList/`
+For task widget source files, map to the widget-specific ai-docs dynamically:
+- `packages/contact-center/task/src/widgets/{WidgetName}/*` → `packages/contact-center/task/ai-docs/widgets/{WidgetName}/` (if the ai-docs folder exists)
+
+This covers all current and future widgets (e.g., CallControl, IncomingTask, OutdialCall, TaskList, and any new widgets added later).
 
 **Deduplicate**: If multiple source files map to the same ai-docs folder, validate that folder only once.
 
@@ -140,14 +139,17 @@ ai-docs folders checked: {list}
 After presenting the validation report (regardless of findings), create a verification marker so the pre-commit hook allows the commit:
 
 ```bash
-# Hash ALL staged contact-center files — must match the hook's hash logic exactly
-STAGED_CC=$(git diff --cached --name-only 2>/dev/null | grep "^packages/contact-center/" | sort)
+# Hash staged content (not just paths) — must match the hook's hash logic exactly
+CC_PKG="packages/contact-center"
+STAGED_CC=$(git diff --cached --name-only 2>/dev/null | grep "^${CC_PKG}/")
 if [ -n "$STAGED_CC" ]; then
-  HASH=$(echo "$STAGED_CC" | shasum | cut -d' ' -f1)
+  HASH=$(git diff --cached -- "$CC_PKG" | (shasum 2>/dev/null || sha256sum) | cut -d' ' -f1)
   touch "/tmp/.spec-drift-verified-${HASH}"
   echo "Verification marker created: /tmp/.spec-drift-verified-${HASH}"
 fi
 ```
+
+> **Note:** The verification marker covers only currently **staged** content. If you modify and re-stage files after verification, the content hash changes and you will need to re-run `/spec-drift-changed`.
 
 Report to the user: "Verification marker created. The pre-commit hook will allow the next commit for these staged files."
 
@@ -155,7 +157,8 @@ Report to the user: "Verification marker created. The pre-commit hook will allow
 
 - Do NOT auto-fix anything — report findings only
 - Always read actual source code to verify — never assume
-- Use the Task tool with `subagent_type: "Explore"` for checker agents
+- Use the Agent tool with `subagent_type: "Explore"` for checker agents
 - Run agents in parallel when multiple folders are affected
-- Always create the verification marker at the end, even if there are findings (the developer decides whether to fix or commit as-is)
-- The marker hash MUST match the hook's hash computation — both hash ALL staged `packages/contact-center/` files sorted alphabetically
+- If an agent does not return within a reasonable time, note it as "Timed out — manual review needed" in the report and continue with available results
+- Always create the verification marker at the end, even if there are findings — this tool is **advisory**: the developer decides whether to fix or commit as-is
+- The marker hash MUST match the hook's hash computation — both use `git diff --cached` content (not just file paths)
--- a/.claude/commands/spec-drift-changed.md
+++ b/.claude/commands/spec-drift-changed.md
@@ -1,0 +1,161 @@
+# Spec Drift Detection — Changed Files Only
+
+Validate ai-docs affected by any staged/unstaged changes — whether the changes are to ai-docs themselves OR to source code that has corresponding ai-docs. Lightweight mode for pre-commit validation.
+
+## Step 1: Find All Changed Files
+
+Run these commands to find all changed files (staged and unstaged):
+
+```bash
+# Get both staged and unstaged changed files
+(git diff --name-only HEAD 2>/dev/null; git diff --name-only --cached 2>/dev/null) | sort -u
+```
+
+## Step 2: Identify ai-docs That Need Validation
+
+From the changed files, build TWO lists:
+
+### List A: Changed ai-docs files
+Filter for files matching `ai-docs/*.md`. These docs changed directly and need validation.
+
+### List B: Source code files with corresponding ai-docs
+For each changed source file under `packages/contact-center/` (excluding ai-docs files themselves):
+1. Walk up the file's directory path
+2. Check if any ancestor directory contains an `ai-docs/` folder
+3. If yes, that `ai-docs/` folder needs validation against the updated source code
+
+Use this to discover ai-docs folders:
+```bash
+find packages/contact-center -type d -name "ai-docs"
+```
+
+Build a mapping like:
+```
+Changed source file                                                → ai-docs to validate
+packages/contact-center/store/src/store.ts                         → packages/contact-center/store/ai-docs/
+packages/contact-center/task/src/widgets/CallControl/CallControl.tsx → packages/contact-center/task/ai-docs/widgets/CallControl/
+packages/contact-center/cc-components/src/SomeComponent.tsx        → packages/contact-center/cc-components/ai-docs/
+```
+
+For task widget source files, map to the widget-specific ai-docs:
+- `packages/contact-center/task/src/widgets/CallControl/*` → `packages/contact-center/task/ai-docs/widgets/CallControl/`
+- `packages/contact-center/task/src/widgets/IncomingTask/*` → `packages/contact-center/task/ai-docs/widgets/IncomingTask/`
+- `packages/contact-center/task/src/widgets/OutdialCall/*` → `packages/contact-center/task/ai-docs/widgets/OutdialCall/`
+- `packages/contact-center/task/src/widgets/TaskList/*` → `packages/contact-center/task/ai-docs/widgets/TaskList/`
+
+**Deduplicate**: If multiple source files map to the same ai-docs folder, validate that folder only once.
+
+### Combine Lists A and B
+The final set of ai-docs folders to validate is the union of:
+- Folders containing files from List A
+- Folders identified from List B
+
+If NEITHER list has entries, report: **"No ai-docs affected by current changes — nothing to check."** and stop.
+
+## Step 3: Validate Affected ai-docs
+
+For each ai-docs folder that needs validation, spawn an Explore agent with this prompt:
+
+```
+You are validating SDD documentation accuracy against source code.
+
+SOURCE OF TRUTH (actual code): {source_code_directory}
+DOCS TO VALIDATE: {ai_docs_folder} (all .md files in this folder)
+
+CHANGED SOURCE FILES (if any): {list of changed source files in this package}
+CHANGED DOC FILES (if any): {list of changed ai-docs files}
+
+Read every markdown file in the ai-docs folder. For each document, check these 7 categories:
+
+1. FILE TREE: Read any documented file/directory trees. Glob the actual directory. Report missing/extra files.
+
+2. METHOD/API SIGNATURES: For every method documented, read the actual source and verify: name, params, param types, return type, modifiers. Flag any mismatch. Pay special attention to methods in changed source files — new methods may be missing from docs, or changed signatures may not be reflected.
+
+3. TYPE DEFINITIONS: For every type/enum/interface documented, find the actual definition in source. Compare name, fields, field types, enum values. Check if changed source files introduced new types not yet documented.
+
+4. EVENT NAMES: For every event constant or observable referenced, verify it exists in source with the exact name. For MobX: verify @observable/@computed/@action decorators. For React: verify prop callback names. Check if changed source files added new events not yet documented.
+
+5. ARCHITECTURE PATTERNS: Verify claims about MobX store patterns, React component patterns, singleton vs factory, component hierarchy, store injection.
+
+6. LINK VALIDATION: For every relative link [text](path), verify the target exists on disk.
+
+7. CODE EXAMPLES: For every code block, verify API names, method names, parameter names, import paths, MobX patterns are correct.
+
+For each finding, report:
+- File: (path)
+- Line/Section: (approximate line or section heading)
+- Category: (1-7)
+- Severity: Blocking / Important / Medium / Minor
+  - Blocking = wrong API that would cause runtime errors if an AI agent follows the docs
+  - Important = wrong params/types that would cause compilation errors
+  - Medium = incomplete or stale info (e.g., new methods/types/events missing from docs)
+  - Minor = broken links, cosmetic issues
+- What doc says: (quoted)
+- What code actually has: (evidence with file:line)
+- Suggested fix: (exact replacement text)
+```
+
+Run all agents in parallel if multiple ai-docs folders are affected.
+
+## Step 4: Consolidate and Report
+
+Present findings in this format:
+
+```markdown
+## Spec Drift Report — Changed Files
+Generated: {date}
+Trigger: {source code changes / ai-docs changes / both}
+ai-docs folders checked: {list}
+
+### Changed Source Files
+{list of changed source files and their mapped ai-docs folder}
+
+### Changed ai-docs Files
+{list of changed ai-docs files, or "None"}
+
+### Summary
+
+| ai-docs Folder | Findings | Blocking | Important | Medium | Minor |
+|----------------|----------|----------|-----------|--------|-------|
+| ...            |          |          |           |        |       |
+
+### Blocking Findings
+...
+
+### Important Findings
+...
+
+### Medium Findings
+...
+
+### Minor Findings
+...
+
+### Actionable Fixes by File
+(grouped by file path, each with exact old text -> new text)
+```
+
+## Step 5: Create Verification Marker
+
+After presenting the validation report (regardless of findings), create a verification marker so the pre-commit hook allows the commit:
+
+```bash
+# Hash ALL staged contact-center files — must match the hook's hash logic exactly
+STAGED_CC=$(git diff --cached --name-only 2>/dev/null | grep "^packages/contact-center/" | sort)
+if [ -n "$STAGED_CC" ]; then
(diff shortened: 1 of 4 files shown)


.claude/hooks/check-ai-docs-drift.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
```
Shreyas281299, line 1: ### HIGH: Hook is not registered in settings.json -- dead code

How will we set this? The hook script exists but the PR does NOT include any changes to `.claude/settings.json` or `.claude/settings.local.json` to register it as a `PreToolUse` hook. The current `.claude/settings.local.json` only contains permissions -- no hooks configuration. Without registration, this hook will never execute and the entire gating mechanism is non-functional.

**Suggested fix:** Add hook registration to `.claude/settings.json`:
```json
{
  "hooks": {
    "PreToolUse": [
      {
        "matcher": "Bash",
        "command": ".claude/hooks/check-ai-docs-drift.sh"
      }
    ]
  }
}
```

.claude/hooks/check-ai-docs-drift.sh
```diff
@@ -0,0 +1,49 @@
...
+#   4. If none -> exit 0 (allow)
+#   5. Check for verification marker (created by /spec-drift-changed)
+#   6. If marker exists -> delete it, exit 0 (allow, one-time pass)
+#   7. If no marker -> exit 2 (BLOCK, instruct to run /spec-drift-changed)
+
+CC_PKG="packages/contact-center"
+
+# Read tool input from stdin (JSON with tool_input.command)
+INPUT=$(cat)
+COMMAND=$(echo "$INPUT" | python3 -c "import sys,json; print(json.load(sys.stdin).get('tool_input',{}).get('command',''))" 2>/dev/null)
```
Shreyas281299, line 25: ### HIGH: Silent failure when `python3` is unavailable

The script uses `python3 -c` to parse JSON from stdin, with `2>/dev/null` suppressing errors. If `python3` is not installed, `COMMAND` silently becomes an empty string, the `case` statement falls through to `*) exit 0`, and the hook always passes -- silently disabling the gate with no warning.

**Suggested fix:** Add a prerequisite check before this line:
```bash
command -v python3 >/dev/null 2>&1 || { echo "python3 required for spec-drift hook" >&2; exit 0; }
```
Alternatively, use `jq` or pure bash string matching for the simpler JSON structure.

.claude/commands/spec-drift-changed.md
```diff
@@ -0,0 +1,161 @@
...
+fi
+```
+
+Report to the user: "Verification marker created. The pre-commit hook will allow the next commit for these staged files."
+
+## Rules
+
+- Do NOT auto-fix anything — report findings only
+- Always read actual source code to verify — never assume
+- Use the Task tool with `subagent_type: "Explore"` for checker agents
```
Shreyas281299: ### HIGH: References non-existent "Task tool" instead of "Agent tool"

While a `TaskCreate` tool exists in Claude Code, it is for task list/progress tracking and does **not** accept a `subagent_type` parameter. The `subagent_type: "Explore"` parameter belongs to the **Agent tool**. Claude would fail or do the wrong thing when trying to follow this instruction.

**Suggested fix:**
```suggestion
- Use the Agent tool with `subagent_type: "Explore"` for checker agents
```

.claude/commands/spec-drift.md
```diff
@@ -0,0 +1,199 @@
...
+packages/contact-center/user-state/ai-docs/                       → packages/contact-center/user-state/src/
+packages/contact-center/ui-logging/ai-docs/                       → packages/contact-center/ui-logging/src/
+packages/contact-center/test-fixtures/ai-docs/                    → packages/contact-center/test-fixtures/src/
+widgets-samples/cc/samples-cc-react-app/ai-docs/                  → widgets-samples/cc/samples-cc-react-app/src/
+... (discover all that exist on the current branch)
+```
+
+## Step 2: Spawn Checker Agents in Parallel
+
+Use the Task tool to spawn agents. **All agents run in parallel.**
```
Shreyas281299: ### HIGH: References non-existent "Task tool" instead of "Agent tool"

Same issue as in `spec-drift-changed.md`. `TaskCreate` is for task tracking, not for spawning agents. The `subagent_type: "Explore"` parameter belongs to the **Agent tool**.

**Suggested fix:**
```suggestion
Use the Agent tool to spawn agents. **All agents run in parallel.**
```

.claude/commands/spec-drift.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Actionable Fixes by File
+(grouped by file path, each with exact old text -> new text)
+```
+
+## Rules
+
+- Do NOT auto-fix anything — report findings only
+- Always read actual source code to verify — never assume
+- Use the Task tool with `subagent_type: "Explore"` for all checker agents
```
Shreyas281299: ### HIGH: Same "Task tool" → "Agent tool" fix needed here

**Suggested fix:**
```suggestion
- Use the Agent tool with `subagent_type: "Explore"` for all checker agents
```

.claude/hooks/check-ai-docs-drift.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Get staged files under the contact-center package
+STAGED_CC=$(git diff --cached --name-only 2>/dev/null | grep "^${CC_PKG}/")
+
+if [ -z "$STAGED_CC" ]; then
+  exit 0  # No contact-center files staged, allow commit
+fi
+
+# Compute hash of all staged contact-center files
+HASH=$(echo "$STAGED_CC" | sort | shasum | cut -d' ' -f1)
```
Shreyas281299: ### (OPTIONAL) MEDIUM: Marker hash is content-agnostic -- allows stale verification bypass

The hash is computed from sorted file **names** only, not file contents. If a developer runs `/spec-drift-changed`, then modifies staged file contents without changing which files are staged, the marker is still valid and the commit proceeds without re-verification.

While the marker is a good-faith advisory signal (not a tamper-proof seal), including content in the hash would be a low-cost improvement.

**Suggested fix:** Include the content diff in the hash:
```suggestion
HASH=$(git diff --cached -- "$CC_PKG" | shasum | cut -d' ' -f1)
```
Apply the same change to the marker creation in `spec-drift-changed.md` Step 5.

.claude/commands/spec-drift-changed.md
```diff
@@ -0,0 +1,161 @@
...
+### Minor Findings
+...
+
+### Actionable Fixes by File
+(grouped by file path, each with exact old text -> new text)
+```
+
+## Step 5: Create Verification Marker
+
+After presenting the validation report (regardless of findings), create a verification marker so the pre-commit hook allows the commit:
```
Shreyas281299, line 139: ### MEDIUM: Staged vs. unstaged marker mismatch may confuse developers

Step 1 discovers changed files from **both** staged and unstaged changes (`git diff --name-only HEAD` + `git diff --name-only --cached`). Steps 2-4 validate against all of these. But the marker in Step 5 only hashes `git diff --cached --name-only` (staged only).

If unstaged changes cause drift and appear in the report, the developer sees them but the marker doesn't cover them. When those changes are later staged, a new verification run is needed -- but the developer may think "I already verified this."

**Suggested fix:** Add a note after this line:
> **Note:** The verification marker covers only currently **staged** files. If you stage additional changes after verification, you will need to re-run `/spec-drift-changed`.

.claude/commands/spec-drift.md
```diff
@@ -0,0 +1,199 @@
...
+(incomplete info, stale file trees — would cause confusion)
+
+### Minor Findings
+(broken links, cosmetic issues)
+
+### Actionable Fixes by File
+(grouped by file path, each with exact old text -> new text)
+```
+
+## Rules
```
Shreyas281299, line 192: ### MEDIUM: No timeout or failure handling for parallel agents

The full scan command spawns potentially 12+ parallel Explore agents but provides no guidance on what happens if an agent times out, fails, or returns empty results. Claude could hang waiting for stuck agents with no fallback.

**Suggested fix:** Add a rule here:
```
- If an agent does not return within a reasonable time, note it as "Timed out -- manual review needed" in the report and continue with available results.
```

.claude/hooks/check-ai-docs-drift.sh
```diff
@@ -0,0 +1,49 @@
...
+MARKER="/tmp/.spec-drift-verified-${HASH}"
```
Shreyas281299, line 48: ### LOW: `/tmp/` marker location is a hardening opportunity

Marker files in `/tmp/` with predictable names are functional for local dev workflows but could be improved. Using a project-local directory (e.g., `.claude/.spec-drift-markers/`) would make the mechanism more self-contained and avoid interaction with shared `/tmp/` cleanup policies.

.claude/hooks/check-ai-docs-drift.sh
```diff
@@ -0,0 +1,49 @@
...
+#   7. If no marker -> exit 2 (BLOCK, instruct to run /spec-drift-changed)
+
+CC_PKG="packages/contact-center"
+
+# Read tool input from stdin (JSON with tool_input.command)
+INPUT=$(cat)
+COMMAND=$(echo "$INPUT" | python3 -c "import sys,json; print(json.load(sys.stdin).get('tool_input',{}).get('command',''))" 2>/dev/null)
+
+# Only gate git commit commands
+case "$COMMAND" in
```
Shreyas281299, line 34: ### LOW: Overly broad `case` pattern for commit detection

The pattern `git\ commit*` matches any command starting with "git commit", including git plumbing commands like `git commit-tree`. While unlikely to cause issues in practice, a more precise pattern would be:
```bash
case "$COMMAND" in
  "git commit"|git\ commit\ *) ;;  # Match 'git commit' standalone or with args
  *) exit 0 ;;
esac
```

.claude/hooks/check-ai-docs-drift.sh
```diff
@@ -0,0 +1,49 @@
...
+STAGED_CC=$(git diff --cached --name-only 2>/dev/null | grep "^${CC_PKG}/")
+
+if [ -z "$STAGED_CC" ]; then
+  exit 0  # No contact-center files staged, allow commit
+fi
+
+# Compute hash of all staged contact-center files
+HASH=$(echo "$STAGED_CC" | sort | shasum | cut -d' ' -f1)
+MARKER="/tmp/.spec-drift-verified-${HASH}"
+
```
Shreyas281299, line 49: ### LOW: `shasum` portability across platforms

`shasum` (SHA-1 default) is macOS-native. On some Linux distributions, only `sha256sum` is available. For cross-platform teams, consider a fallback:
```bash
HASH=$(echo "$STAGED_CC" | sort | (shasum 2>/dev/null || sha256sum) | cut -d' ' -f1)
```

.claude/commands/spec-drift-changed.md
```diff
@@ -0,0 +1,161 @@
...
+Build a mapping like:
+```
+Changed source file                                                → ai-docs to validate
+packages/contact-center/store/src/store.ts                         → packages/contact-center/store/ai-docs/
+packages/contact-center/task/src/widgets/CallControl/CallControl.tsx → packages/contact-center/task/ai-docs/widgets/CallControl/
+packages/contact-center/cc-components/src/SomeComponent.tsx        → packages/contact-center/cc-components/ai-docs/
+```
+
+For task widget source files, map to the widget-specific ai-docs:
+- `packages/contact-center/task/src/widgets/CallControl/*` → `packages/contact-center/task/ai-docs/widgets/CallControl/`
```
Shreyas281299: ### LOW: Hardcoded widget path mappings

The mapping for task widget source files to ai-docs is hardcoded to 4 specific widgets (CallControl, IncomingTask, OutdialCall, TaskList). New widgets added to `task/src/widgets/` would not be automatically detected.

**Suggested fix:** Add a dynamic pattern instruction:
```
For any widget under packages/contact-center/task/src/widgets/{WidgetName}/*,
map to packages/contact-center/task/ai-docs/widgets/{WidgetName}/ if the
ai-docs folder exists.
```

.claude/commands/spec-drift-changed.md
```diff
@@ -0,0 +1,161 @@
...
+
+Report to the user: "Verification marker created. The pre-commit hook will allow the next commit for these staged files."
+
+## Rules
+
+- Do NOT auto-fix anything — report findings only
+- Always read actual source code to verify — never assume
+- Use the Task tool with `subagent_type: "Explore"` for checker agents
+- Run agents in parallel when multiple folders are affected
+- Always create the verification marker at the end, even if there are findings (the developer decides whether to fix or commit as-is)
```
Shreyas281299: ### LOW: Marker created regardless of blocking findings (design observation)

This rule makes the tool purely advisory -- even "Blocking" severity findings (described as "wrong API that would cause runtime errors") won't actually block the commit. Consider documenting this advisory-only nature more prominently in the report output, so developers understand the tool's stance.

.claude/commands/spec-drift.md
```diff
@@ -0,0 +1,199 @@
...
+
+1. **Repo detection**: This is ccWidgets — `packages/contact-center/` exists (no `@webex` scope)
+2. **Root AGENTS.md**: `AGENTS.md` (repo root)
+3. **Framework docs**: `ai-docs/` (README, RULES, patterns/*, templates/*)
+4. **Package-level ai-docs**: Glob for `packages/contact-center/**/ai-docs/` to find all ai-docs folders
+5. **Samples ai-docs**: Check `widgets-samples/**/ai-docs/` as well
+
+For each ai-docs folder found, identify its corresponding source code directory (the parent directory of `ai-docs/`).
+
+Build an inventory like:
```
Shreyas281299: ### LOW: Hardcoded ai-docs inventory is exemplary but could be clearer

The inventory table lists 12+ specific paths and ends with `"... (discover all that exist on the current branch)"`. While clearly intended as illustrative, consider adding an explicit label like `"Example inventory (actual results may vary):"` to prevent any ambiguity.

Shreyas281299 (changes requested):
## Review Summary

| Severity | Count |
|----------|-------|
| High | 3 |
| Medium | 3 |
| Low | 6 |

**Recommendation: Request Changes** -- The 3 High severity issues should be addressed before merge:
1. Register the hook in `settings.json` (otherwise the hook is dead code)
2. Fix "Task tool" -> "Agent tool" references (otherwise the commands will fail)
3. Add `python3` availability check (otherwise the hook silently becomes a no-op)

### Positive Observations
- Well-structured command files with clear step-by-step instructions and output format templates
- The 7-category drift detection framework is comprehensive and well-defined
- Good use of severity levels (Blocking/Important/Medium/Minor) in the drift report
- The incremental scan (`spec-drift-changed`) is a practical addition for pre-commit workflows
- The shell hook logic is concise and handles the common cases correctly

.claude/hooks/check-ai-docs-drift.sh
ciscoRankush: Fixed in 3cf0a3f6. Added `.claude/settings.json` with the PreToolUse hook registration for Bash commands.

.claude/hooks/check-ai-docs-drift.sh
ciscoRankush: Fixed in 3cf0a3f6. Added `python3` availability check (`command -v python3`) and a fail-closed guard — if `COMMAND` is empty after parsing, the hook now exits with code 2 (block) instead of 0 (allow).

.claude/commands/spec-drift-changed.md
ciscoRankush: Fixed in 3cf0a3f6. Changed all references from "Task tool" to "Agent tool" with `subagent_type: "Explore"`.

.claude/commands/spec-drift.md
ciscoRankush: Fixed in 3cf0a3f6. Changed to "Agent tool" here as well.

.claude/commands/spec-drift.md
ciscoRankush: Fixed in 3cf0a3f6. Updated to "Agent tool" in the Rules section as well.

.claude/hooks/check-ai-docs-drift.sh
ciscoRankush: Fixed in 3cf0a3f6. Hash now uses `git diff --cached -- "$CC_PKG"` (staged content) instead of just file paths. Also added `shasum || sha256sum` fallback. The matching hash logic in `spec-drift-changed.md` Step 5 was updated to match.

.claude/commands/spec-drift-changed.md
ciscoRankush: Fixed in 3cf0a3f6. Added a note after the marker creation: "The verification marker covers only currently **staged** content. If you modify and re-stage files after verification, the content hash changes and you will need to re-run `/spec-drift-changed`."

.claude/commands/spec-drift.md
ciscoRankush: Fixed in 3cf0a3f6. Added rule: "If an agent does not return within a reasonable time, note it as 'Timed out — manual review needed' in the report and continue with available results."

.claude/hooks/check-ai-docs-drift.sh
ciscoRankush: Acknowledged. Keeping `/tmp/` for now as it works for local dev workflows and avoids adding project-local directories that need gitignore rules. Can revisit if we see issues with shared cleanup policies.

.claude/hooks/check-ai-docs-drift.sh
ciscoRankush: Fixed in 3cf0a3f6. Updated case pattern to `"git commit"|git\ commit\ *` to avoid matching `git commit-tree` and similar plumbing commands.

.claude/hooks/check-ai-docs-drift.sh
ciscoRankush: Fixed in 3cf0a3f6. Added `(shasum 2>/dev/null || sha256sum)` fallback in both the hook and the marker creation in `spec-drift-changed.md`.

.claude/commands/spec-drift-changed.md
ciscoRankush: Fixed in 3cf0a3f6. Replaced the 4 hardcoded widget paths with a dynamic pattern: `packages/contact-center/task/src/widgets/{WidgetName}/*` → `packages/contact-center/task/ai-docs/widgets/{WidgetName}/` (if ai-docs folder exists). New widgets are automatically covered.

.claude/commands/spec-drift-changed.md
ciscoRankush: Fixed in 3cf0a3f6. Updated the rule to explicitly state: "this tool is **advisory**: the developer decides whether to fix or commit as-is".

.claude/commands/spec-drift.md
ciscoRankush: Fixed in 3cf0a3f6. Changed to "Build an inventory (example — actual results will vary based on current branch):".

Shreyas281299 (approved):

.claude/commands/spec-drift-changed.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Step 2: Identify ai-docs That Need Validation
+
+From the changed files, build TWO lists:
+
+### List A: Changed ai-docs files
+Filter for files matching `ai-docs/*.md`. These docs changed directly and need validation.
+
+### List B: Source code files with corresponding ai-docs
+For each changed source file under `packages/contact-center/` (excluding ai-docs files themselves):
```
mkesavan13, line 22: I believe we have the specs done for meetings as well. So, do we need to keep it only for contact center or for both?

.claude/commands/spec-drift-changed.md
```diff
@@ -0,0 +1,164 @@
...
+
+Report to the user: "Verification marker created. The pre-commit hook will allow the next commit for these staged files."
+
+## Rules
+
+- Do NOT auto-fix anything — report findings only
+- Always read actual source code to verify — never assume
+- Use the Agent tool with `subagent_type: "Explore"` for checker agents
+- Run agents in parallel when multiple folders are affected
+- If an agent does not return within a reasonable time, note it as "Timed out — manual review needed" in the report and continue with available results
```
mkesavan13, line 162: What if the LLM decides to do a long reasonable wait time?

mkesavan13 (approved):
Looks good. Just a couple of minor comments. Particularly whether we want to keep it only for contact center or extend it to the Meetings Widget as well because we have specs there now.

.claude/commands/spec-drift-changed.md
ciscoRankush: we should not do here , there should be a separate skill for meetings , as this skill specific to cc

.claude/commands/spec-drift-changed.md
ciscoRankush: but still we need to put some threshhold at some pint of time . lets see how its behaving , and start observing

PR #690, opened 2026-05-28
fix(store):  attach SDK listeners on stationLoginSuccess

vamshigovardhana:
# COMPLETES https://jira-eng-sjc12.cisco.com/jira/browse/CAI-7899

UserState (and other CC widgets that depend on `agent:stateChange`) freeze in non-WebRTC tabs — e.g. Epic Hyperspace secondary windows that intentionally skip Mobius registration, or any Extension/AgentDN login where Mobius is not registered.

Root cause: `addEventListeners()` in `storeEventsWrapper.setupIncomingTaskHandler` was only invoked from the `AGENT_DN_REGISTERED` and `AGENT_RELOGIN_SUCCESS` handlers. When Mobius registration is skipped and silent relogin does not run, neither event fires, so the `agent:stateChange` (and related) listeners are never attached and the store never receives state updates from the SDK.

- Call `addEventListeners()` from the `AGENT_STATION_LOGIN_SUCCESS` handler (`handleLogin`) as well, guarded by the same `listenersAdded` flag so listeners are attached exactly once regardless of which event fires first.

Builds on the sample-app changes from #683 which add a "Disable WebRTC Registration" toggle for repro convenience.

### Change Type

- [x] Bug fix (non-breaking change which fixes an issue)

- Extension login (Mobius skipped) → `agent:stateChange` listener is attached on `stationLoginSuccess`; UserState updates propagate (no freeze).
- Browser login with Mobius registered → listener attached exactly once (no duplicate via `agent:dnRegistered`).
- Silent relogin path (`agent:reloginSuccess`) → listener attached exactly once.

## The GAI Coding Policy And Copyright Annotation Best Practices

- [x] GAI was used to create a draft that was subsequently customized
- [x] Tool used for AI assistance
  - [x] Other - Please Specify: Cursor (Claude)
- [x] This PR is related to
  - [x] Defect fix

### Checklist before merging

- [ ] I have not skipped any automated checks
- [ ] All existing and new tests passed
- [ ] I have updated the testing document
- [ ] I have tested the functionality with amplify link

### DIFF
--- a/packages/contact-center/store/src/storeEventsWrapper.ts
+++ b/packages/contact-center/store/src/storeEventsWrapper.ts
@@ -923,6 +923,13 @@ class StoreWrapper implements IStoreWrapper {
         // @ts-expect-error To be fixed in SDK - https://jira-eng-sjc12.cisco.com/jira/browse/CAI-6762
         this.setTeamId(payload.teamId);
       });
+      // CAI-7899: Ensure CC SDK listeners (incl. agent:stateChange) are attached even when
+      // Mobius registration is skipped (non-WebRTC tabs e.g. Epic Hyperspace secondary windows,
+      // or Extension/AgentDN logins) and silent relogin doesn't run.
+      if (!listenersAdded) {
+        addEventListeners();
+        listenersAdded = true;
+      }
     };
 
     ccSDK.on(CC_EVENTS.AGENT_STATION_LOGIN_SUCCESS, handleLogin);
--- a/widgets-samples/cc/samples-cc-react-app/src/App.tsx
+++ b/widgets-samples/cc/samples-cc-react-app/src/App.tsx
@@ -90,6 +90,14 @@ function App() {
     const savedAllowInternationalDn = window.localStorage.getItem('allowInternationalDn');
     return savedAllowInternationalDn === 'true';
   });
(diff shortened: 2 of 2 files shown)


packages/contact-center/store/src/storeEventsWrapper.ts
```diff
@@ -1082,6 +923,13 @@ class StoreWrapper implements IStoreWrapper {
         // @ts-expect-error To be fixed in SDK - https://jira-eng-sjc12.cisco.com/jira/browse/CAI-6762
         this.setTeamId(payload.teamId);
       });
+      // CAI-7899: Ensure CC SDK listeners (incl. agent:stateChange) are attached even when
```
Shreyas281299, line 926: We dont need to put the CAI ID in code unless its a TODO

Shreyas281299:
AFAIK, in this scenario we would received AGENT_RELOGIN event and that should add the event listeners. Can you please share a testing video for this fix

Shreyas281299 (changes requested):
Please add test cases and share a testing video for the fix

Shreyas281299:
Ive tested this user-state works fine even without this- the custom states are not visbile in multi-login because thats how its has been designed we have to fix the custom states.

PR #697, opened 2026-06-11
chore(ci): gate docs publish on next only

rsarika:

This follow-up PR brings the same `publish-documentation` guard onto `next` so the deploy workflow behavior is aligned there as well.

- Added `if: github.ref_name == 'next'` to the `publish-documentation` job in `.github/workflows/deploy.yml`
- Kept the rest of the deploy workflow unchanged

## Notes

- This is a small follow-up PR against `next`
- It mirrors the docs-publish guard introduced for `task-refactor`-based work so deployment behavior is consistent once merged

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches: # Allow list of deployable tags and branches. Note that all allow-listed branches cannot include any `/` characters
       - next
+      - task-refactor
 
 env:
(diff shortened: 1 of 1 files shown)


Shreyas281299 (approved):

rsarika:
Superseded by #698 so the follow-up branch uses the plain name `stop-gh-pages-task-refactor-next` instead of a `codex/` prefix.

